PR #7, opened 2016-05-21
Logging improvements for testing

n1rvana:


n1rvana:
These changes are developer tested and showed no regressions from m2f.

PR #17, opened 2016-08-02
factom-cli receipt EntryHash get the earliest entry merkletree

Eleven711:
actom-cli receipt EntryHash show the earliest entry merkletree


carryforward:
tested and merged

PR #18, opened 2016-08-12
# fixed "factom-cli get alltxs" am/pm error

Eleven711:


carryforward:
committed to m2-monday

PR #21, opened 2016-08-31
add authentication connection on factomd server side

Eleven711:
add factomd -rpcuser -rpcpassword
factomd, factom-walletd, factom-cli rpcuser and rpcpassword should be same
they can be configed in config file


carryforward:
merged into 11 branch
https://github.com/FactomProject/factomd/tree/11

PR #22, opened 2016-09-01
Add CheckNetworkID

Eleven711:
add checknetworkid


carryforward:
added here: https://github.com/FactomProject/factomd/blob/19d441b82d0a0a1030a0068cf5578c7cd062f9a5/common/messages/dbstate.go#L148

PR #24, opened 2016-09-06
fix factom-walletd flage -w error

Eleven711:
fix factom-walletd flage -w error


carryforward:
the problem turned out to be that a directory was being passed in, and bolt can't change a folder to a file.  not it checks if it is a folder first.
https://github.com/FactomProject/factom/commit/a2bbaffddfff397b46e43ae2e51ffa962acfb2d1

PR #48, opened 2016-12-16
Fixed the search for and request for missing entries and entryblocks

PaulSnow:
JIRA-340

Big stability fix.  Fixes entry chain and entry transactions.

### DIFF
--- a/state/state.go
+++ b/state/state.go
@@ -1108,6 +1108,10 @@ func (s *State) NoEntryYet(entryhash interfaces.IHash, ts interfaces.Timestamp) 
 	return unique
 }
 
+// This routine walks through the directory blocks, looking for missing entry blocks and entries.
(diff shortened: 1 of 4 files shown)


state/state.go
```diff
@@ -1158,34 +1156,17 @@ func (s *State) catchupEBlocks() {
...
-						doubleCheck = true
-					}
-				}
-			} else {
-				doubleCheck = true
-			}
 
-			if doubleCheck {
+			if dbstate != nil {
 				db := s.GetDirectoryBlockByHeight(s.EntryBlockDBHeightProcessing)
```
dieswaytoofast: Formatting (then again, this isn't python :-)  )

dieswaytoofast:
General commentary - the module is *huge*, and could do w/ breaking up.  But @ThePiachu has already pointed that out :-)

Else, LGTM

PR #51, opened 2016-12-16
Sof 415  Fix to only ignore messages once at boot, even if the network delays.  

PaulSnow:
This is actually just a small update.  Rolled into this is also a fix to GetHighestSavedBlk() and GetHighestCompletedBlk() which were reversed in actual function.  This change is just a rename of one for the other in the code.

### DIFF
--- a/state/stateConsensus.go
+++ b/state/stateConsensus.go
@@ -87,8 +87,11 @@ func (s *State) Process() (progress bool) {
 		if now-s.StartDelay > s.StartDelayLimit {
 			if s.DBFinished == true {
 				s.RunLeader = true
-				s.StartDelay = now // Reset StartDelay for Ignore Missing
-				s.IgnoreMissing = true
+				if !s.IgnoreDone {
+					s.StartDelay = now // Reset StartDelay for Ignore Missing
+					s.IgnoreMissing = true
+					s.IgnoreDone = true
+				}
 			}
 		}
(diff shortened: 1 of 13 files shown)


state/stateConsensus.go
```diff
@@ -87,8 +87,11 @@ func (s *State) Process() (progress bool) {
 		if now-s.StartDelay > s.StartDelayLimit {
 			if s.DBFinished == true {
 				s.RunLeader = true
-				s.StartDelay = now // Reset StartDelay for Ignore Missing
-				s.IgnoreMissing = true
+				if !s.IgnoreDone {
+					s.StartDelay = now // Reset StartDelay for Ignore Missing
```
dieswaytoofast: Just to be _completely_ paranoid, if `IgnoreDone` is `true`, then `StartDelay` does *not* get reset, correct?

hannemennah: Correct. IgnoreDone is not initialized to a particular value _except_ right here, though (so it is initially going to be false as things boot up). And if IgnoreDone _is_ true, then that should indicate that we don't need to reset the state's StartDelay value (since we don't need to go through the "Ignore interval"; that phase should be done).

hannemennah:
Other than the observation Mahesh made, everything else in this PR is just function renaming (which needed to be done; the old names were misleading).

PR #52, opened 2016-12-17
Sof 399:  Wait to the EOM before we save to disk

PaulSnow:
Avoids issues with backing out the database when blocks are not shared among the leaders.


dieswaytoofast:
LGTM

PR #55, opened 2016-12-18
Checkpoint of better working dbstate handling

PaulSnow:
More work to fix issues with changing the dbstate handling.

Added debugging messages and logging in many places.

Did some cleanup related to logging.

Removed faulting messages that are out of date from holding.


dieswaytoofast:
LGTM

PR #58, opened 2016-12-19
Sof 399

PaulSnow:
Pulled in stability changes in sof-456 into this branch.

Better test script.  Now testing entry sizes from 100's of bytes to 9900 bytes.

Applied the fix to FullFault messages that we have already applied ot acks and msgs when missing.


dieswaytoofast:
LGTM

PR #63, opened 2016-12-19
Fixed test

PaulSnow:
Not sure why the code expected 10 blocks.  If we wait long enough to have 10 blocks, the unit tests will take forever.  Also the genesis block has changed, so the hash had to be updated.


dieswaytoofast:
LGTM

PR #66, opened 2016-12-20
Sof 463

hannemennah:


common/messages/dbstate.go
```diff
@@ -185,9 +185,36 @@ func (m *DBStateMsg) Validate(state interfaces.IState) int {
...
+	validSigCount := 0
+
+	data, err := m.DirectoryBlock.GetHeader().MarshalBinary()
+	if err != nil {
+		state.AddStatus(fmt.Sprint("Debug: DBState Signature Error, Marshal binary errored"))
+		return validSigCount
+	}
+
+	for _, sig := range m.SignatureList.List {
+		check, err := state.VerifyAuthoritySignature(data, sig.GetSignature(), dbheight)
```
PaulSnow: How do we tell the difference between audit and federated servers here?

state/state.go
```diff
@@ -926,7 +927,33 @@ func (s *State) LoadDBState(dbheight uint32) (interfaces.IMsg, error) {
 		panic(fmt.Sprintf("The configured network ID (%x) differs from the one in the local database (%x) at height %d", configuredID, dbaseID, dbheight))
 	}
 
-	msg := messages.NewDBStateMsg(s.GetTimestamp(), dblk, ablk, fblk, ecblk, eBlocks, entries)
+	blockSig := new(primitives.Signature)
+	var allSigs []interfaces.IFullSignature
+
+	nextABlock, err := s.DB.FetchABlockByHeight(dbheight + 1)
+	if err != nil || nextABlock == nil {
+		return nil, err
```
PaulSnow: This is where you can look in the process list to see if you can get the signatures there.

hannemennah:
Updated to also check processLists for the signatures.

state/state.go
```diff
@@ -932,27 +932,36 @@ func (s *State) LoadDBState(dbheight uint32) (interfaces.IMsg, error) {
...
 	nextABlock, err := s.DB.FetchABlockByHeight(dbheight + 1)
 	if err != nil || nextABlock == nil {
-		return nil, err
-	}
-	abEntries := nextABlock.GetABEntries()
-	for _, adminEntry := range abEntries {
-		data, err := adminEntry.MarshalBinary()
-		if err != nil {
-			return nil, err
+		if dbheight < s.ProcessLists.DBHeightBase || int(dbheight) > int(s.ProcessLists.DBHeightBase)+len(s.ProcessLists.Lists) {
```
PaulSnow: This test isn't necessary.  You can just get the process list, and if it is nil, then you don't have one.

PR #68, opened 2016-12-20
Sof 439

mwanon:
avoided re-adding pending entries when already in response

added IsInPendingEntryList to factomd/util/misc.go 

added a couple nil checks the are part of other SOF tasks, but arent in yet.

got rid of s.ProcessLists.LastList() in favor of iterating through last 2 which works better

### DIFF
--- a/state/state.go
+++ b/state/state.go
@@ -1064,6 +1064,7 @@ func (s *State) fillAcksMap() {
 }
 
 func (s *State) GetPendingEntries(params interface{}) []interfaces.IPendingEntry {
+	fmt.Println("GetPendingEntries")
 	resp := make([]interfaces.IPendingEntry, 0)
 	pls := s.ProcessLists.Lists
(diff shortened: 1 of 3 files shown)


state/state.go
```diff
@@ -1106,65 +1107,73 @@ func (s *State) GetPendingEntries(params interface{}) []interfaces.IPendingEntry
...
-						enb, err := plmsg.MarshalBinary()
-						if err != nil {
-							return nil
-						}
-						err = cc.UnmarshalBinary(enb)
-						if err != nil {
-							return nil
-						}
-						tmp.EntryHash = cc.CommitChain.EntryHash
+		if pls != nil {
```
n1rvana: Shouldn't this be:
if pl != nil?
if pls is nil, what does range pls do?

n1rvana: Line 1217 below seems to confirm my theory.

PR #82, opened 2016-12-22
Sof 490

PaulSnow:
Clean up to throw away old messages that avoids bogus heartbeat conflict errors (and other conflict errors)


dieswaytoofast:
LGTM

PR #91, opened 2016-12-24
Fix  for holding entries when building identities

PaulSnow:


dieswaytoofast:
Works, per @VeenaGondkar

PR #100, opened 2016-12-28
Sof 504

PaulSnow:
Updated glide.yaml, fixed a test script


PaulSnow:
Invalid.  Missing revered changes.  One more coming...

PR #102, opened 2016-12-28
Fixed lock file, updated all.sh to use glide

PaulSnow:
Made all.sh script pull appropriate branches from git, then glide them.

Updated the lock file for factomd


dieswaytoofast:
LGTM

PR #104, opened 2016-12-29
check for nil

PaulSnow:
I keep putting this in, and it keeps getting lost in the shuffle. 

Ignore the nil process list, return, and process later when the ducks are actually in order.


dieswaytoofast:
LGTM
(All four lines)
(Including the blank line)

PR #106, opened 2016-12-30
Sof 509

PaulSnow:
Improved download management of dbstates

### DIFF
--- a/state/dbStateManager.go
+++ b/state/dbStateManager.go
@@ -164,7 +164,7 @@ func (ds *DBState) String() string {
 		str = "  Directory Block = <nil>\n"
 	} else {
 
-		str = fmt.Sprintf("%s      State: IsNew %5v ReadyToSave %5v Locked %5v Signed %5v Saved %5v", str, ds.IsNew, ds.ReadyToSave, ds.Locked, ds.Signed, ds.Saved)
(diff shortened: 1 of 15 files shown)


state/dbStateManager.go
```diff
@@ -642,12 +646,10 @@ func (list *DBStateList) SignDB(d *DBState) (process bool) {
 }
 
 func (list *DBStateList) SaveDBStateToDB(d *DBState) (progress bool) {
+	dbheight := int(d.DirectoryBlock.GetHeader().GetDBHeight())
 	// Take the height, and some function of the identity chain, and use that to decide to trim.  That
 	// way, not all nodes in a simulation Trim() at the same time.
-	v := int(d.DirectoryBlock.GetHeader().GetDBHeight()) + int(list.State.IdentityChainID.Bytes()[0])
```
Emyrk: list.State.IdentityChainID.Bytes()[0] is usually 0x88, should be any byte >4

state/dbStateManager.go
```diff
@@ -664,6 +666,12 @@ func (list *DBStateList) SaveDBStateToDB(d *DBState) (progress bool) {
 		return
 	}
 
+	// Only trim when we are really saving.
+	v := dbheight + int(list.State.IdentityChainID.Bytes()[0])
```
Emyrk: list.State.IdentityChainID.Bytes()[0] is usually 0x88, should be any byte >4

Emyrk:
My only concern is the only loading of 200 processlists at a time may be too slow. It does address the memory consumption, but slows down the loading. Unsure of a good number to compromise

PR #110, opened 2016-12-31
SOF-514: network buffering

gornik:
After some investigation it looks like gob has a bit of a inconsistent behavior when sending/receiving data:

* in the decoder, the reader is wrapped in a bufio.Reader instance:
  https://github.com/golang/go/blob/master/src/encoding/gob/decoder.go#L39-L42
* the encoder writes everything to the writer in one shot:
  https://github.com/golang/go/blob/master/src/encoding/gob/encoder.go#L81

This means that if when we call `SetReadDeadline` on `c.conn`, the timeout applies to reading a single buffer from the connection (4kB by default), while when we call `SetWriteDeadline` on `c.conn`, the timeout applies to the entire message.

This PR wraps the `c.conn` in the `bufio.Reader` and `bufio.Writer` instances, so that both reads and writes are buffered before sending and the timeouts apply only to a single buffer. The size of the buffer is set in `NetworkReadBufferSize` and `NetworkWriteBufferSize` (10kB right now).

Buffering on the read size should not be necessary (`gob.Decoder` already does that), but I've added it so we have the control over the size of the read buffer.

### DIFF
--- a/p2p/protocol.go
+++ b/p2p/protocol.go
@@ -49,6 +49,8 @@ var (
 	MinumumSharingQualityScore    int32  = 20          // if a peer's score is less than this we don't share them.
 	OnlySpecialPeers                     = false
 	NetworkDeadline                      = time.Duration(25000) * time.Millisecond
+	NetworkReadBufferSize                = (1024 * 10) // 10kB, size of the read buffer for data sent over the network
+	NetworkWriteBufferSize               = (1024 * 10) // 10kB, size of the write buffer for data sent over the network
 	NumberPeersToConnect                 = 8
 	MaxNumberIncommingConnections        = 150
 	MaxNumberOfRedialAttempts            = 15
--- a/p2p/connection.go
+++ b/p2p/connection.go
@@ -5,6 +5,7 @@
 package p2p
 
 import (
+	"bufio"
 	"bytes"
 	"encoding/gob"
 	"fmt"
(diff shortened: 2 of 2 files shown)


p2p/protocol.go
```diff
@@ -49,6 +49,8 @@ var (
 	MinumumSharingQualityScore    int32  = 20          // if a peer's score is less than this we don't share them.
 	OnlySpecialPeers                     = false
 	NetworkDeadline                      = time.Duration(25000) * time.Millisecond
+	NetworkReadBufferSize                = (1024 * 10) // 10kB, size of the read buffer for data sent over the network
+	NetworkWriteBufferSize               = (1024 * 10) // 10kB, size of the write buffer for data sent over the network
```
dieswaytoofast: A good exercise would be to actually set this to 1K (i.e., eliminate both MTUs and TCP-segment-size from the equation).

p2p/protocol.go
```diff
@@ -49,6 +49,8 @@ var (
 	MinumumSharingQualityScore    int32  = 20          // if a peer's score is less than this we don't share them.
 	OnlySpecialPeers                     = false
 	NetworkDeadline                      = time.Duration(25000) * time.Millisecond
+	NetworkReadBufferSize                = (1024 * 10) // 10kB, size of the read buffer for data sent over the network
```
dieswaytoofast: Same as Write Buffer, i.e., set to 1K

PR #112, opened 2016-12-31
SOF-514: Recreate encoder/decoder when sending data

gornik:
Recreate encoder/decoder on each write/read:
- on writes: recreate the encoder and encode to a buffer first, then send encoded data in chunks over the socket
- on reads: whenever we receive some data, try to decode it using a new decoder, if we have an error when decoding and it is not a recognized network error, discard bad data, log and continue

### DIFF
--- a/p2p/connection.go
+++ b/p2p/connection.go
@@ -5,6 +5,7 @@
 package p2p
 
 import (
+	"bufio"
 	"bytes"
 	"encoding/gob"
 	"fmt"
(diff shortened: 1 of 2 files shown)


p2p/connection.go
```diff
@@ -460,30 +479,63 @@ func (c *Connection) processReceives() {
 	for ConnectionOnline == c.state {
 		var message Parcel
 		verbose(c.peer.PeerIdent(), "Connection.processReceives() called. State: %s", c.ConnectionState())
+
+		var bufReader = bufio.NewReaderSize(c.conn, NetworkReadBufferSize)
+
+		b, _ := bufReader.Peek(2048)
```
dieswaytoofast: Is this for deugging purposes?

gornik: Yes, this is only temporary, so we take a look at what's causing the decoder to crash.

PR #116, opened 2017-01-01
Sof 525

PaulSnow:
This line can cause a nil pointer.  It isn't even necessary.


hannemennah:
LGTM (just a nil sanity check)

PR #124, opened 2017-01-03
SOF-514: parts mixup fix

gornik:
* Moves partial parcel assembly from connection to the controller (also separated the code to a separate file)
* Allows parts to arrive in random order, the final full message is assembled according to the AppHash from the original message
* If the expected parts stop arriving, cleans the existing entries after min 10 minutes (the check is performed every time a part arrives)
* Deacreases payload size from 10kB to 1kB to avoid gob errors

### DIFF
--- a/p2p/parts_assembler.go
+++ b/p2p/parts_assembler.go
@@ -19,23 +19,23 @@ type PartialMessage struct {
 
 // PartsAssembler is responsible for assembling message parts into full messages
 type PartsAssembler struct {
-	messages map[uint64]*PartialMessage // a map of full message IDs to
+	messages map[string]*PartialMessage // a map of app hashes to partial messages
 }
 
 // Initializes the assembler
 func (assembler *PartsAssembler) Init() *PartsAssembler {
-	assembler.messages = make(map[uint64]*PartialMessage)
+	assembler.messages = make(map[string]*PartialMessage)
 	return assembler
 }
 
 // Handles a single message part, returns either a fully assembled message or nil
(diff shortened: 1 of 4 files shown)


p2p/parts_assembler.go
```diff
@@ -0,0 +1,87 @@
...
+
+	if !exists {
+		partial = createNewPartialMessage(parcel)
+		assembler.messages[parcel.Header.AppHash] = partial
+	}
+
+	partial.parts[parcel.Header.PartNo] = &parcel
+	partial.mostRecentPartReceived = time.Now()
+
+	fullParcel := tryReassemblingMessage(partial)
```
dieswaytoofast: For legibility, perhaps a comment here that `tryReassemblingMessage` returns the message or `nil`
(to map to the overall comment on L31)
It will avoid having to do a lookup in the code to understand this...

p2p/parts_assembler.go
```diff
@@ -0,0 +1,87 @@
...
+	if fullParcel != nil {
+		delete(assembler.messages, parcel.Header.AppHash)
+		debug("PartsAssembler", "Fully assembled %d", parcel.Header.AppHash)
+	}
+
+	assembler.cleanupOldPartialMessages()
```
dieswaytoofast: Comment here pointing out that we do Garbage Collection on _each received part_

p2p/parts_assembler.go
gornik: Comment added

p2p/parts_assembler.go
gornik: Comment added

PR #125, opened 2017-01-05
SOF-533: Validate parcel part before assembling

gornik:
Invalid parcel parts are now rejected


hannemennah:
This is definitely a needed change. I approve.

We may need to extend validateParcelPart() to include other validation cases later on down the line, but this is without a doubt a step in the right direction.

PR #127, opened 2017-01-05
Catch panics during parcel-part validation/handling

hannemennah:


p2p/connection.go
```diff
@@ -504,6 +504,12 @@ func (c *Connection) handleNetErrors(err error) {
 // handleParcel checks the parcel command type, and either generates a response, or passes it along.
 // return value:  Indicate whether we got a good message or not and thus whether we should keep reading from network
 func (c *Connection) handleParcel(parcel Parcel) {
+	defer func() {
+		if r := recover(); r != nil {
+			return
```
n1rvana: I think a panic is worth logging.  I would add the following  after checking recover()
c.peer.demerit() /// so someone DDoS or just incompatible will eventually be cut off after 200+ panics
            fmt.Fprintf(os.Stdout, "Caught Exception in connection %s: %v\n", c.peer.PeerFixedIdent(), err)

otherwise LGTM

PR #135, opened 2017-01-07
Add back deleted code to keep EBDBHeightComplete updated properly

hannemennah:


dieswaytoofast:
@PaulSnow - this one is in your court, since [it seems to revert your earlier patch to entrySyncing.go](https://github.com/FactomProject/factomd/pull/133/files#diff-d6dd78a5e753e403090588da84c9851cL98)

PR #139, opened 2017-01-09
Fix build, add coveralls support

gornik:
* Ignores vendor dir when unit tests are run
* Integrates coveralls to ensure unit test coverage
* Adds a `.coverignore` file that lets you specify which files should be excluded from coverage calculations, the lines in this file can contain any wildcards supported by https://golang.org/pkg/path/filepath/#Match

### DIFF
--- a/.coverignore
+++ b/.coverignore
@@ -1,0 +1,2 @@
+controlPanel/files/statics/*
+controlPanel/files/templates/*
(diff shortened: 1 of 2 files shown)


Emyrk:
The test to see if the code is good is if all automated checks pass... they do!

PR #148, opened 2017-01-14
Improved load times, trimmed dbstates down.

PaulSnow:
Appears to greatly improve network performance and load times.

Tested in simulation, and tested with production data on an Azure server.

Mostly involves tweaking parameters.


PaulSnow:
Updated to a different branch

PR #155, opened 2017-01-24
Fixed a small issue with EntryServerFault

ThePiachu:
It looks like EntryServerFault wasn't marshalling its type.


ThePiachu:
re: https://factom.atlassian.net/browse/SOF-561

PR #162, opened 2017-01-31
Some logic cleanup to prevent panics in simcontrol

Emyrk:
Also some excess code line cleanup in the sim control.


PaulSnow:
This code only impacts sim control and avoids a run time panic.  Steven is the primary user.

PR #168, opened 2017-02-07
update databases

carryforward:
This fixes many problems, including the too many files issue, the windows bolt issue.


Emyrk:
The failing of the tests is not caused by this commit. This is an exisiting problem. Still going to merge this as the database updates are needed.

PR #169, opened 2017-02-07
enable compression for LevelDB

carryforward:
This reduces the database size from 9 gb to 4 gb after syncing the blockchain.
it also affects the factom-walletd block cache when its glide is updated.


carryforward:
this should be pulled in after the leveldb is updated to the latest version in https://github.com/FactomProject/factomd/pull/168

PR #173, opened 2017-02-27
Sof 606 into develop

Emyrk:
Fixed control panel showing invalid addresses as valid
Simple search 'FA3eQZMJBVpsTKRB19x9tC1wD3fzwSi4Q7Y5BW32rYw6P5yyWqfrExtracharacters' to see if it has been fixed. Also made it break out of the switch and go through the other searches if the address search fails. This is because EC and FA are valid hex, so there could be a case where an entry with "FA..." or "EC..." get's improperly rejected. Now it is fixed.


carryforward:
it works on my machine.

PR #178, opened 2017-03-10
Sof 602

ThePiachu:


Utilities/DatabasePorter/api.go
```diff
@@ -130,7 +130,7 @@ func GetEntry(hash string) (interfaces.IEBEntry, error) {
 		}
 		return entry, nil
```
PaulSnow: There is no loop here.... You can't try 10 times because you will always return, right?

Utilities/DatabasePorter/api.go
```diff
@@ -130,7 +130,7 @@ func GetEntry(hash string) (interfaces.IEBEntry, error) {
...
 	}
-	panic("Failed 100 times to get the data " + hash)
+	//panic("Failed 100 times to get the data " + hash)
 	return nil, nil
 }
```
PaulSnow: This is dead code.

Utilities/DatabasePorter/api.go
ThePiachu: You only return if there is no error. If there is an error, you will continue and skip the return.

Utilities/DatabasePorter/api.go
ThePiachu: No, if you keep getting errors, you will keep doing `continue` until you run out of the loop and then you return nil nil

PR #182, opened 2017-03-17
Move dbhash and keymr to top of cli output for 'get dbheight' calls

hannemennah:


Emyrk:
LGTM -- Tested with factom-cli

PR #184, opened 2017-03-17
Add txids and useraddresses to tx json output

hannemennah:


Emyrk:
User Addresses are all lowercase by this function in the WSAPI. This function sets all fields in the json body to lowercase, but it also catches all content.

https://github.com/FactomProject/factomd/blob/SOF-563/wsapi/wsapiV2.go#L344

hannemennah:
Had to manually tag the json fields to get the keys but not the values. It's done now.

PR #188, opened 2017-03-22
remove double processing of messages

PaulSnow:
You will note that we execute the message on line 171.  Then (if the message is in process) again on about line 196.  That can make us process a dbstate message twice.


carryforward:
yay, big bug fix.

PR #202, opened 2017-04-06
Fd 32

PaulSnow:


Emyrk:
Fixed the route to random peer. Was selecting first in map each time before.

Everything else looks good

PR #204, opened 2017-04-10
Exposed a debug API for authorities

Emyrk:
Added api call to get authority set. 
Made structs have json tags


carryforward:
looks good to me.

PR #210, opened 2017-04-12
Added an API call to get transaction rate

Emyrk:
Moved the calculation for transaction rate to a loop that is guarenteed
execution, and made it execute every ~3 seconds. This is a decent estimate
of transactions seen by the node


dieswaytoofast:
Instrumentation?

Emyrk:
Added Prometheus instrumentation

michaelbeam:
Looks great. Good job @Emyrk

PR #214, opened 2017-04-13
Fd 50

ThePiachu:


PaulSnow:
I wonder if buffer shouldn't panic, and make the marshal catch the error if they want to report one.

Cuts 3 lines to 1 per element marshaled.

PaulSnow:
Aged out at this point

PR #240, opened 2017-04-28
Instrument fix

Emyrk:
BroadCastIn Queue drops messages when it is full. These were not counted in the Prometheus, meaning the gauge will not return to 0 if messages are dropped. This was fixed, and another Prometheus counter was added to count the dropped messages


dieswaytoofast:
LGTM

PR #242, opened 2017-04-28
Fd 87

PaulSnow:
Adds constants where we had hard coded values for limits.
Adds limits to missing data request handling.


ThePiachu:
/engine unit tests appear to be failing, please fix.

PaulSnow:
What a pain.  There was a timing issue with the code (a check for a fast follow for add2).  Have to pause.

PR #280, opened 2017-06-27
Change to api.  Added current-minute call with associated changes to …

mwanon:
…state, etc


mwanon:
changed base from master to develop

PR #287, opened 2017-07-07
Fd 179

Emyrk:
New `ack` api call

### DIFF
--- a/state/stateConsensus.go
+++ b/state/stateConsensus.go
@@ -911,8 +911,14 @@ func (s *State) LeaderExecuteDBSig(m interfaces.IMsg) {
 }
 
 func (s *State) LeaderExecuteCommitChain(m interfaces.IMsg) {
-	s.LeaderExecute(m)
 	cc := m.(*messages.CommitChainMsg)
+	// Check if this commit has more entry credits than any previous that we have.
+	if !s.isHighestCommit(cc.CommitChain.EntryHash, m) {
+		// This commit is not higher than any previous, so we can discard it and prevent a double spend
+		return
+	}
+
+	s.LeaderExecute(m)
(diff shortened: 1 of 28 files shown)


state/stateConsensus.go
```diff
@@ -1884,9 +1896,27 @@ func (s *State) GetNewEBlocks(dbheight uint32, hash interfaces.IHash) interfaces
 	return nil
 }
 
+func (s *State) IsNewOrPendingEBlocks(dbheight uint32, hash interfaces.IHash) bool {
+	if dbheight <= s.GetHighestSavedBlk()+2 {
+		pl := s.ProcessLists.Get(dbheight)
```
hannemennah: Seems like this should be s.ProcessLists.GetSafe(dbheight) instead.

state/stateConsensus.go
```diff
@@ -1896,23 +1931,33 @@ func (s *State) PutNewEntries(dbheight uint32, hash interfaces.IHash, e interfac
...
+	// the highest.
 	switch {
-	case ok1 && ok1b && e.CommitEntry.Credits > m.CommitEntry.Credits:
-	case ok2 && ok2b && ec.CommitChain.Credits > mc.CommitEntry.Credits:
+	case ok1 && ok1b && e.CommitEntry.Credits >= m.CommitEntry.Credits:
+	case ok2 && ok2b && ec.CommitChain.Credits >= mc.CommitChain.Credits:
 	default:
-		s.Commits[hash.Fixed()] = msg
+		// Returns true when new commit is greater than old, or if old does not exist
+		return true
```
hannemennah: Won't this _always_ return true? Seems like the switch is written so that no matter what, "return true" is executed.

hannemennah: Ignore me, I was mentally inserting "fallthrough" here without realizing it.

PR #340, opened 2017-09-21
Set the DBHeight in the Factoid State, debugging, slight optimization…

carryforward:
… in processing EOM, and go fmt


carryforward:
FD-280

PR #348, opened 2017-10-05
FD-281

hannemennah:
Enable Logstash-compatible logging (to integrate with the Elastic stack).


carryforward:
github is not updating when pushes are made to this branch.  recreating pull request.

PR #357, opened 2017-10-18
Fd 292

PaulSnow:
more security around Factoid Transactions.


common/constants/constants.go
```diff
@@ -56,12 +56,15 @@ const (
...
 	NETWORK_REPLAY  = 2
-	TIME_TEST       = 4  // Checks the time_stamp;  Don't put actual hashes into the map with this.
-	REVEAL_REPLAY   = 8  // Checks for Reveal Entry Replays ... No duplicate Entries within our 4 hours!
-	BLOCK_REPLAY    = 16 // Ensures we don't add the same transaction to multiple blocks.
+	TIME_TEST       = 4 // Checks the time_stamp;  Don't put actual hashes into the map with this.
+	REVEAL_REPLAY   = 8 // Checks for Reveal Entry Replays ... No duplicate Entries within our 4 hours!
+
+	// FReplay -- Block based Replay filter consttructed by processing the blocks, from the database
+	//            then from blocks either passed to a node, or constructed by messages.
+	BLOCK_REPLAY = 1 // Ensures we don't add the same transaction to multiple blocks.
```
carryforward: changing a bitfield bits because block replay is a different variable than the message replay detector

common/messages/MessageBase.go
```diff
@@ -125,7 +125,7 @@ func (m *MessageBase) Expire(s interfaces.IState) (rtn bool) {
 	if m.expire == 0 {
 		m.expire = now
 	}
-	if now-m.expire > 5*60*1000 { // Keep messages for some length before giving up.
+	if now-m.expire > 60*60*1000 { // Keep messages for some length before giving up.
```
carryforward: keep messages around for an hour instead of 5 minutes so messages aren't dropped during network segmentation.

state/dbStateManager.go
```diff
@@ -1336,15 +1349,6 @@ func (list *DBStateList) SaveDBStateToDB(d *DBState) (progress bool) {
 		panic(err.Error())
 	}
 
-	// Set the Block Replay flag for all these transactions we are saving to the database.
-	for _, fct := range d.FactoidBlock.GetTransactions() {
-		list.State.FReplay.IsTSValid_(
```
carryforward: This code was moved around to catch when loading from blocks and downloading from messages

state/factoidstate.go
```diff
@@ -293,17 +293,27 @@ func (fs *FactoidState) UpdateECTransaction(rt bool, trans interfaces.IECBlockEn
...
+	}
+	// Then update the state for all inputs.
+	for _, input := range trans.GetInputs() {
+		adr := input.GetAddress().Fixed()
+		oldv := fs.State.GetF(rt, adr)
+		v := oldv - int64(input.GetAmount())
 		fs.State.PutF(rt, adr, v)
 	}
+	// Then log that the transaction has been seen and processed.
+	fs.State.Replay.IsTSValid(constants.INTERNAL_REPLAY, trans.GetSigHash(), trans.GetTimestamp())
```
carryforward: moved to not give false positives when different error paths are exercised.  This gets rid of false positives.

PR #359, opened 2017-10-19
Fd 304

ThePiachu:


Emyrk:
Given the age of this PR, and the fact we don't have this ticket internally anymore, I am going to close it.

PR #363, opened 2017-10-20
Fix Typo

ianwelsh:


carryforward:
Thank you for the bug report.

We don't merge updates directly into master, as we always test things before they get released.  We did include your update in the latest release though.

https://github.com/FactomProject/factomd/commit/c61b18e899c25d21a737959611aa091263fa724c

I am closing this pull request since your fix has been included through the normal channels.

PR #366, opened 2017-10-23
Fd 66

ThePiachu:
Changed DBOverlay to handle block heads and block-by-height using data from DBlocks.


Cahl-Dee:
Determined we are not doing this

PR #373, opened 2017-11-09
Fix random nil pointer errors

PaulSnow:


carryforward:
https://github.com/FactomProject/factomd/commits/FD-308-322

PR #381, opened 2017-11-14
Add Quickstart guide

logicethos:
1. I have added a Quickstart guide. Simple steps to start a node on main net.
2. I note that port 8108 has not been opened in the rest of the document.  This appears to limit the number of main net connections.
3. I would suggest a direct link to the Quickstart instructions is added to https://hub.docker.com/r/factominc/factomd/
4. I added -i to keep STDIN open.  This may not be needed with the recent change to Dockerfile.

### DIFF
--- a/Docker.md
+++ b/Docker.md
@@ -2,6 +2,23 @@
 
 The factomd Docker Helper is a simple tool to help build and run factomd as a container
 
+## Quickstart Guide (public repository)
+
+#### To Run Main Net:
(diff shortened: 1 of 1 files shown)


logicethos:
Further to 4.  I added comments here https://github.com/FactomProject/factomd/commit/9b6a60af3cd53864dbeb65a8afc7a9432894c959

stackdump:
we have this in various other places now

PR #387, opened 2017-11-28
FD-334 Cache marshaled message so we don't repeatedly marshal again

PaulSnow:
Messages passed to a node (the normal case) are unmarshaled to process and validate.  If acceptable, they are then sent broadcast out to the network.

We now keep the marshaled message to avoid the additional memory allocations and marshaling.


PaulSnow:
Close... Dumb.  Wrong branch.

PR #388, opened 2017-12-02
Fd 336

PaulSnow:
Fix an issue where sometimes when we process missing messages, we fail to handle entry reveals correctly.


PaulSnow:
ignore.  Merging the combination of FD-334 and FD-336

PR #389, opened 2017-12-07
FD-334  Optimization 

PaulSnow:
Don't need to marshal messages if we keep their unmarshaled form when we receive them from the network.  Saves Garbage collection, and repeated marshal operations.


PaulSnow:
ignore.  Merging the combination of FD-334 and FD-336

PR #390, opened 2017-12-08
The merge of FD-334 and FD-336

PaulSnow:
Fixes Commit Reveals when network is loaded.

Optimization that reduces footprint and increases performance.


carryforward:
This combines both https://github.com/FactomProject/factomd/pull/389 and https://github.com/FactomProject/factomd/pull/388

PR #395, opened 2018-01-04
Docker-compose based dev environment

gornik:
See added README.md


gornik:
@factom-clay as discussed: merging this one and later I'll replace the peer list passed via the command line with an nginx instance serving the initial peer list.

PR #409, opened 2018-01-24
Fd 347

factom-clay:
Debug Console and stdout/stderr log


factomd.go
```diff
@@ -8,18 +8,237 @@ import (
...
+		os.Exit(1)
+	}
+
+	// launch debug console if requested
+	if (params.DebugConsole != "") {
+		launchDebugServer(params.DebugConsole)
+	}
+
+	//  Go Optimizations...
+	runtime.GOMAXPROCS(runtime.NumCPU()) // TODO: should be *2 to use hyperthreadding? -- clay
```
Emyrk: GOMAXPROCS defaults to the number of cpus. Not sure what it does in a hyper threading situation, but we shouldn't need to change it.

engine/factomParams.go
```diff
@@ -110,7 +112,9 @@ func (f *FactomParams) Init() {
 	f.Sim_Stdin = true
 	f.exposeProfiling = false
 	f.Sync2 = -1
-	f.DebugConsole = false
+	f.DebugConsole = "foobar" //TODO: pretty sure this value is overridden by the default in the flag -- clay
```
Emyrk: It is overwritten by the default in the ParseCmdLine. We should remove this comment. As to why we set defaults here? Can't say for certain...

Emyrk:
Last commit by me was done by Clay on my machine to add stdout to the output stream.

LG

PR #431, opened 2018-03-02
promote v0.4.2.20 from develop to master

carryforward:


PaulSnow:
Note that the drop in coverage (-24 percent) isn't actually real.  The coverage reports have changed in some way, and some of the tools we are using to add up coverage are not picking up on the reporting changes.  We expect this to be resolved in coming weeks.

Don't worry!  We remain committed to testing and verification of our code!

PR #432, opened 2018-03-13
Support/dev env changes

gornik:
Updated the env in `support/dev`:
 - removed starting `factomd` instances - this is all handled by the network testing tool (`support/net`) now and this setup is supposed to only start all the supporting services
 - fixed a race condition during startup (one container starting when another is not ready)
 - added dynamic discovery of services monitored by Prometheus
 - updated the docs, removed irrelevant parts


gornik:
merged into `rollup_062718`

PR #451, opened 2018-04-10
Fix errors after gob enc/dec were reinintialized

gornik:
Fixes two separate issues:

* When the connection was going offline due to some error, it was
possible in some cases for the goOnline function to reinitialize gob
encoder/decoder without restarting the TCP/IP connection. The new
decoder would start reading data from the old connection, but due to
buffered reads the old decoder might have left the data in the
connection in the middle on the message. This caused an error which
again caused the connection to go offline and online and repeat the
problem, so as a result the whole connection was stuck in an infinite
error loop. This change ensures that the connection is closed when going
offline.

* The gob.Decoder.Decode function returns io.EOF when there is nothing
to read. This wasn't handled properly, so it was treated as an error (it
might have been the original cause for the previous problem).  This
change ignores the io.EOF result from the decoder.

### DIFF
--- a/p2p/connection.go
+++ b/p2p/connection.go
@@ -8,6 +8,7 @@ import (
 	"encoding/gob"
 	"fmt"
 	"hash/crc32"
+	"io"
 	"net"
 	"os"
 	"time"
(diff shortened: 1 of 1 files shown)


gornik:
merged into `rollup_062718`

PR #462, opened 2018-05-02
FD-472

Emyrk:
Revert leaderheight to saved block height by leader

Behavior was taking the processlist being built. Not the saved height.
This is the reverting the behavior to that prior to the v5.0.0 release


chrislgarry:
What testing has been done for this change? Can you paste any results here?

chrislgarry:
@drobertbaker just curious, you guys dont paste test results in the PRs (or screenshots where test results arent applicable)?

drobertbaker:
@chrislgarry We have an internal ticketing system for advancing issues from concept through development to QA and merging (JIRA). There we keep all relevant samples,screenshots, attachments, comments, qualifications, results, etc to form a history.

Issue #463, opened 2018-05-03
p2p/peer.go/Peer.LocationFromAddress() panics when Peer.Address contains an interface name as returned by Linux 4.14.36


AdamSLevy:
When Peer.Address contains an address like "fe80::f6:ecff:fe05:5ffc%eth0", Peer.LocationFromAddress panics due to `index out of range`. 

The full panic is:
```
panic: runtime error: index out of range
goroutine 52 [running]:
github.com/FactomProject/factomd/p2p.(*Peer).LocationFromAddress(0xc4200b9a00, 0x15)
        /build/factomd/src/src/github.com/FactomProject/factomd/p2p/peer.go:141 +0x489
github.com/FactomProject/factomd/p2p.(*Peer).Init(0xc4200b9a00, 0xc420130c9b, 0x15, 0xc420130cb1, 0x4, 0x100000000, 0x0, 0x0)
        /build/factomd/src/src/github.com/FactomProject/factomd/p2p/peer.go:65 +0x333
github.com/FactomProject/factomd/p2p.(*Controller).parseSpecialPeers(0xc4202046e0, 0xc420130c80, 0x35, 0x1, 0x44bea8, 0xab72a411278d, 0x800000000c321a26)
        /build/factomd/src/src/github.com/FactomProject/factomd/p2p/controller.go:386 +0x2b6
github.com/FactomProject/factomd/p2p.(*Controller).initSpecialPeers(0xc4202046e0, 0xcbbd22, 0x6, 0xc4201a6130, 0x4, 0xc42001cf30, 0x2b, 0x100feedbeef, 0xc420238150, 0x60, ...)
        /build/factomd/src/src/github.com/FactomProject/factomd/p2p/controller.go:363 +0x78
github.com/FactomProject/factomd/p2p.(*Controller).Init(0xc4202046e0, 0xcbbd22, 0x6, 0xc4201a6130, 0x4, 0xc42001cf30, 0x2b, 0x100feedbeef, 0xc420238150, 0x60, ...)
        /build/factomd/src/src/github.com/FactomProject/factomd/p2p/controller.go:190 +0x819
github.com/FactomProject/factomd/engine.NetStart(0xc4200f0000, 0x12d09c0, 0x0)
        /build/factomd/src/src/github.com/FactomProject/factomd/engine/NetStart.go:378 +0x3f3f
created by github.com/FactomProject/factomd/engine.Factomd
[...]

PR #464, opened 2018-05-03
Fix LocationFromAddress panic on bad Peer.Address

AdamSLevy:
This commit fixes Issue #463. An edge case on Linux concerning how
net.LookupAddress resolves its own hostname results in a panic. Check if
ip is nil once more before indexing into ip inside of
Peer.LocationFromAddress.

https://github.com/FactomProject/factomd/issues/463


AdamSLevy:
rebased on master

stackdump:
We have a change in a4 to log panics - so if this is still occurring we'll re-visit

PR #473, opened 2018-05-11
Fd 492

factom-clay:
Added pull back on highestAsk to make process re ask for nil inserted or created by timing. Changed default faulttimeout to 120


factom-clay:
Handed by future merges that are now in place.

PR #478, opened 2018-05-14
Supresses MMR process while loading from disk, removed trimming the PL when doing elections. fixed sending revels when processing commits.

factom-clay:


stackdump:
changes made it in via other PR's

PR #481, opened 2018-05-16
Fd 497

Emyrk:
Entry syncing and Message Relaying fixes


Emyrk:
This is no longer an issue

PR #485, opened 2018-05-24
Fd 507 concurrent access nillist

factom-clay:
Fixed a concurrent map access when logging in sim with multiple node. Also fixes the setting of HighestAsks. Think this is already in rollup_fixes


factom-clay:
Superseded by other changes

PR #494, opened 2018-06-07
FD-523 Preserve log level

carryforward:


Emyrk:
#495 Achieves this

carryforward:
done in FD-519

PR #511, opened 2018-07-18
Add CORS domains option to allow calls to Factomd API from browsers

PaulBernier:
Dependency on https://github.com/FactomProject/web/pull/2
Pass along CORS domains parameters to allow calls to API from browsers.
That's pretty much my first Goland code ever so I'm open to any suggestion.


stackdump:
cors change made it in via another route

PR #518, opened 2018-07-30
revert from 5.3.1 to 5.2.0 Then apply grants on top of 5.2.0

carryforward:


carryforward:
should be to develop, not master

PR #541, opened 2018-08-28
v5.4.3 Release candidate 1

carryforward:
includes:
FD-587
FD-594
FD-603
FD-606
FD-609
FD-610
FD-618
FD-620
FD-628
FD-630
FD-631
FD-632


carryforward:
merge is pending bake time on testnet.

carryforward:
testnet has baked and is ready for release

PR #576, opened 2018-09-19
Fix unallocated map bug

factom-clay:


carryforward:
this was pulled into master in another branch

PR #577, opened 2018-09-20
fix issue /w memory on fast boot

stackdump:
Setting DBHeight Base fixed the memory spike issue.

Found that when we no longer load a process list for all blocks, this caused the db height in the Header to be incorrect so added code to correct this when it happens.

TODO: still need to work on adding a unit test for this

### DIFF
--- a/state/processList.go
+++ b/state/processList.go
@@ -1166,9 +1166,15 @@ func NewProcessList(state interfaces.IState, previous *ProcessList, dbheight uin
 		pl.AdminBlock = adminBlock.NewAdminBlock(previous.AdminBlock)
 		pl.EntryCreditBlock, err = entryCreditBlock.NextECBlock(previous.EntryCreditBlock)
 	} else {
-		pl.DirectoryBlock = directoryBlock.NewDirectoryBlock(nil)
-		pl.AdminBlock = adminBlock.NewAdminBlock(nil)
-		pl.EntryCreditBlock, err = entryCreditBlock.NextECBlock(nil)
+		if pl.DBHeight > 0 {
+			pl.DirectoryBlock, _ = state.GetDB().FetchDBlockByHeight(pl.DBHeight)
+			pl.AdminBlock, _ = state.GetDB().FetchABlockByHeight(pl.DBHeight)
+			pl.EntryCreditBlock, _ = state.GetDB().FetchECBlockByHeight(pl.DBHeight)
+		} else {
+			pl.DirectoryBlock = directoryBlock.NewDirectoryBlock(nil)
+			pl.AdminBlock = adminBlock.NewAdminBlock(nil)
+			pl.EntryCreditBlock, err = entryCreditBlock.NextECBlock(nil)
+		}
 	}
 
 	pl.ResetDiffSigTally()
--- a/state/saveAndRestore.go
+++ b/state/saveAndRestore.go
@@ -574,6 +574,9 @@ func (ss *SaveState) RestoreFactomdState(s *State) { //, d *DBState) {
 	}
 	// Set this, as we know it to be true
 	s.DBHeightAtBoot = ss.DBHeight
+
+	// set DBHeightBase to avoid recursively loading process lists
+	s.ProcessLists.DBHeightBase = ss.DBHeight
 	pl := s.ProcessLists.Get(ss.DBHeight)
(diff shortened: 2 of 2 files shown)


state/processList.go
```diff
@@ -1169,6 +1169,11 @@ func NewProcessList(state interfaces.IState, previous *ProcessList, dbheight uin
 		pl.DirectoryBlock = directoryBlock.NewDirectoryBlock(nil)
 		pl.AdminBlock = adminBlock.NewAdminBlock(nil)
 		pl.EntryCreditBlock, err = entryCreditBlock.NextECBlock(nil)
+
```
stackdump: Wondering if there are any other side effects like this for the Admin & Entry Credit Blocks

stackdump: @PaulSnow I think there may be a few options - should we store more information about these Blocks in the Fastboot save file or try to retrieve db entries to repopulate

state/processList.go
```diff
@@ -1166,9 +1166,15 @@ func NewProcessList(state interfaces.IState, previous *ProcessList, dbheight uin
...
-		pl.EntryCreditBlock, err = entryCreditBlock.NextECBlock(nil)
+		if pl.DBHeight > 0 {
+			pl.DirectoryBlock, _ = state.GetDB().FetchDBlockByHeight(pl.DBHeight)
+			pl.AdminBlock, _ = state.GetDB().FetchABlockByHeight(pl.DBHeight)
+			pl.EntryCreditBlock, _ = state.GetDB().FetchECBlockByHeight(pl.DBHeight)
+		} else {
+			pl.DirectoryBlock = directoryBlock.NewDirectoryBlock(nil)
+			pl.AdminBlock = adminBlock.NewAdminBlock(nil)
+			pl.EntryCreditBlock, err = entryCreditBlock.NextECBlock(nil)
+		}
```
stackdump: it seems like maybe I should add a change like this
```
diff --git a/state/processList.go b/state/processList.go
index da9866957..73e3f471f 100644
--- a/state/processList.go
+++ b/state/processList.go
@@ -1167,9 +1167,13 @@ func NewProcessList(state interfaces.IState, previous *ProcessList, dbheight uin
                pl.EntryCreditBlock, err = entryCreditBlock.NextECBlock(previous.EntryCreditBlock)
        } else {
                if pl.DBHeight > 0 {
-                       pl.DirectoryBlock, _ = state.GetDB().FetchDBlockByHeight(pl.DBHeight)
-                       pl.AdminBlock, _ = state.GetDB().FetchABlockByHeight(pl.DBHeight)
-                       pl.EntryCreditBlock, _ = state.GetDB().FetchECBlockByHeight(pl.DBHeight)
+                       prevDBlock, _ := state.GetDB().FetchDBlockByHeight(pl.DBHeight-1)
+                       prevAblock, _ := state.GetDB().FetchABlockByHeight(pl.DBHeight-1)
+                       prevECBlock, _ := state.GetDB().FetchECBlockByHeight(pl.DBHeight-1)
+
+                       pl.DirectoryBlock = directoryBlock.NewDirectoryBlock(prevDBlock)
+                       pl.AdminBlock = adminBlock.NewAdminBlock(prevAblock)
+                       pl.EntryCreditBlock, err = entryCreditBlock.NextECBlock(prevECBlock)
                } else {
                        pl.DirectoryBlock = directoryBlock.NewDirectoryBlock(nil)
                        pl.AdminBlock = adminBlock.NewAdminBlock(nil)

```
My worry is that these blocks have already been stored - whereas generally Process Lists seem to only hold un-commited db blocks

state/saveAndRestore.go
```diff
@@ -574,6 +574,9 @@ func (ss *SaveState) RestoreFactomdState(s *State) { //, d *DBState) {
 	}
 	// Set this, as we know it to be true
 	s.DBHeightAtBoot = ss.DBHeight
+
+	// set DBHeightBase to avoid recursively loading process lists
+	s.ProcessLists.DBHeightBase = ss.DBHeight
 	pl := s.ProcessLists.Get(ss.DBHeight)
```
stackdump: If we added additional change  w/ noted in comments to ProcessList.go - it may be that this should change to
```
	pl := s.ProcessLists.Get(ss.DBHeight+1)
```

PR #581, opened 2018-09-27
added memleak fix by not cloning Replay during save

stackdump:
This change may have some unintended consequences in regard to what information is persisted from Replay - need to talk about that w/ @factom-clay 

NOTE: I wonder if could rebuild Replay data from DB rather than persisting to the state snapshot

### DIFF
--- a/state/replay.go
+++ b/state/replay.go
@@ -96,6 +96,11 @@ func (a *Replay) IsSameAs(b *Replay) bool {
 }
 
 func (r *Replay) MarshalBinary() (rval []byte, err error) {
+	// REVIEW: added locking since Marshal is being executed
+	// against State.Replay rather than a copy from Replay.Save()
+	r.Mutex.Lock()
+	defer r.Mutex.Unlock()
+
 	defer func(pe *error) {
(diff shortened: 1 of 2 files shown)


stackdump:
Event with this fix mem usage still grows from ~ 4GB to over 7GB in about an hour
Running profiler again - next largest offender is SaveDBStateList

stackdump:
breaks this test & others https://github.com/FactomProject/factomd/blob/2daca5356ec8c9ee08f26fc9e4147613593b9937/common/messages/dbstate_test.go#L220

state/replay.go
```diff
@@ -153,6 +158,9 @@ func (r *Replay) UnmarshalBinary(p []byte) error {
 }
 
 func (r *Replay) Save() *Replay {
+	// KLUDGE don't make a copy to prevent memory leak
```
stackdump: This caused some test failures - but did 'fix' the mem lean - though there is still a leak that seems to be from SaveDBStateList()

PaulSnow: likely not what we want...

PR #596, opened 2018-10-12
Add Paul Bernier

PaulBernier:


carryforward:
Thank you Paul!  Glad to have the first non-Factom, Inc code contribution!

PR #598, opened 2018-10-18
FD-654 Add CORS and update many libraries

carryforward:
This code was created by @PaulBernier even though it is showing as being merged in by JoshuaBrigati.

The glide.yaml was fixed to include the CORS library, as well as freshly recreating the dependent repos based on th current state of the codebase.  The glide.lock file also includes a year of updates from libraries factomd uses.

Thank you Luap for writing this code to allow javascript browser access to the factomd API, and for the patience waiting for the patch to work its way to production.


PaulBernier:
Thanks for reviewing my PR and considering merging it! I did a quick sanity check and it seems in order.

PR #607, opened 2018-11-13
Fd 740 smart unmarshal limits

michaelbeam:
create limits at runtime based on the size of the binary structures being unmarshaled to stop bad length and size values from allocating arbitrarily large amounts of memory.


common/factoid/fblock.go
```diff
@@ -321,23 +322,34 @@ func (b *FBlock) UnmarshalBinaryData(data []byte) ([]byte, error) {
 		return nil, err
 	}
 
-	cnt, err := buf.PopUInt32()
+	// txLimit is the maximum number of transactions (min 78 bytes) that could
+	// fit into the remaining unread portion of the buffer.
+	txLimit := uint32(buf.Len() / 78)
```
sambarnes: Swap ordering of limit and count assignments (?unsure about if this one really matters tho..)

common/identity/entryBlockSync.go
```diff
@@ -136,14 +136,24 @@ func (e *EntryBlockSync) UnmarshalBinaryData(p []byte) (newData []byte, err erro
 		return
 	}
 
-	l, err := buf.PopInt()
+	// blockLimit is the mazimum number of Entry Blocks that could fit in the
+	// buffer. Smallest possible Entry Block is 140 bytes.
+	blockLimit := buf.Len() / 140
```
sambarnes: Swap ordering of limit and count assignments (?again, unsure if this one would actually be affected, or if the division already covers it enough. pointing out just in case)

common/adminBlock/EntryAddEfficiency_test.go
```diff
@@ -38,3 +38,20 @@ func TestAddEfficiency(t *testing.T) {
...
 	}
 }
+
+func TestAddBadEfficiency(t *testing.T) {
+	e1 := NewAddEfficiency(primitives.RandomHash(), uint16(5000))
+	p, err := e1.MarshalBinary()
+	if err != nil {
+		t.Error(err)
+	}
+	p[1] = 0xff // replace body lenght with bad value
```
michaelbeam:
```suggestion
	p[1] = 0xff // replace body length with bad value
```

common/adminBlock/EntryAddEfficiency.go
```diff
@@ -123,33 +123,41 @@ func (e *AddEfficiency) UnmarshalBinaryData(data []byte) ([]byte, error) {
...
 		return nil, err
 	}
 
-	if b != e.Type() {
+	if t != e.Type() {
 		return nil, fmt.Errorf("Invalid Entry type")
 	}
 
-	bl, err := buf.PopVarInt()
+	bodyLimit := uint64(buf.Len())
```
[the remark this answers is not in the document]
michaelbeam: counting the VarInt as part of the size is generally negligible, and it will only ever slightly increase the limit, which dosn't hinder the purpose of the new limits

Issue #608, opened 2018-11-14
Show hostname on console and control panel


stackdump:
From Discord: 

[Bedrock] jcheroskeToday at 12:57 PM
Not sure if this is the right place to ask this, but I was wondering if the 8090 console could have the hostname added to the display? All my nodes look alike!
...
[Bedrock] jcheroskeToday at 1:33 PM
It's obviously not a big deal, but it's a feature I'd like to see in the console. That way, no matter how you connect to your node, you'll see the name that the node itself is called. Seems useful

sambarnes:
The enterprise wallet does the same thing, and this is it's explanation:
> This compiles our HTML, CSS, JS, and images into golang .go files. That way we don't need to worry about pathing while serving files as everything is compiled into the binary.

To compile the modified templates and js for the control panel, there's some instructions here:
https://github.com/FactomProject/factomd/blob/master/controlPanel/README.md

This section of the codebase is mostly from Steven I  believe. I'm not all that familiar with it yet, I just know the basic gist of it from when I was looking at enterprise-wallet stuff

stackdump:
FYI currently would be OK to branch from Master to work on this - when you are ready someone can create an FD-XXX branch for you to merge into

Emyrk:
@PaulBernier 
The compile script is here: https://github.com/FactomProject/factomd/blob/master/controlPanel/compile.sh
This is the dependency for building: https://github.com/FactomProject/staticfiles

I need to update the readme, but we decided to not minify things, as the extra build step wasn't buying us much, as most of the files are served over localhost. We can move back to a minified build as the control panels are being used more remotely, but for now the 'uncompressed' part of that readme is out of date.

PaulBernier:
Ok I managed to make the code change to display the info in the control panel this way:

![screenshot from 2018-11-17 11-26-25](https://user-images.githubusercontent.com/5141652/48653795-c1413600-ea5b-11e8-9cdc-66b13774ef80.png)

I asked jcheroske and operators and Discord and it was fine.
The title mentions display in console too, what place is this referring to exactly?

Now that works fine if factond is running directly on the OS... but I experimented with the docker container and the hostname of a docker container is the container id (not the hostname of the host OS)... Maybe there is an option to force the hostname of the container, would need to be investigated.

PaulBernier:
[...]

PR #620, opened 2018-11-25
Add a flag -nodename to assign a custom name display in control panel…

PaulBernier:
… and console
 
For issue  #608

For instance running the following cmd on my laptop:
`$ factomd -network=LOCAL -nodename=$(hostname)`

screenshot from 2018-11-25 19-08-05
screenshot from 2018-11-25 19-07-58


carryforward:
oh, wow, this is fantastic.  Thank you so much @PaulBernier.  I'll pass this onto robert and veena to test.  I might have to rebuild the binary files to confirm correctness, but this is going to be very useful going forward.  I'm not sure what release this will make it into.  There is so much piling up into the next release.  This might be small enough to fit in though.

carryforward:
ok, it turns out the compilation is deterministic, so the audit of the binaries checked out ok.

https://github.com/PaulBernier/factomd/blob/03898df79c0f213f080cfd27da5247960463f15d/controlPanel/README.md

It is ready to test!

PR #621, opened 2018-11-27
Fd 706 move perm blance to save2

PaulSnow:
Don't update the perm balances until we actually save to disk, to avoid issues where we update our current state with a db-state message after we have processed the data collected in a block.


factom-clay:
Merged in mainline

PR #623, opened 2018-12-10
Fd 761 sim add fnodes 

stackdump:
added ./simTest folder w/ a new test that restores a sim using fastsave

NOTE: currently this test shares the same DB as fnode0 - will continue to correct this as part of another PR


stackdump:
replaced with https://github.com/FactomProject/factomd/pull/682

Issue #643, opened 2019-01-30
FD-819 CrossReplayFilter garbage collection never ends


WhoSoup:
https://github.com/FactomProject/factomd/blob/0b30f390e9650f01df12e6f8b3a7384f0f1761f5/state/crossBootReplay.go#L137

This was added to stop the CrossReplay Garbage Collector after ten minutes (to match the other code), but this condition ("now < bootTime - 10 minutes") never evaluates to true, so it keeps running every 5 seconds for eternity. 

I don't think it causes any issues apart from taking up a slight amount of resources.

(this is the counterpart, which appropriately stops after 10 minutes: https://github.com/FactomProject/factomd/blob/0b30f390e9650f01df12e6f8b3a7384f0f1761f5/engine/NetworkProcessorNet.go#L121)

PR #644, opened 2019-01-30
crossreplayfilter garbage collection termination fix

WhoSoup:
pull request for issue https://github.com/FactomProject/factomd/issues/643

Changed to calculate the time of completion once when initialized, then checking if that time has passed


carryforward:
ok, this can get reviewed in our development process.  this is kind of low priority, but it is a good catch.  Thank you for the update.

PR #647, opened 2019-02-14
Fd 831 legibility improvements batch 1

WhoSoup:
there's one bugfix in there that might change behavior, the other commits should not change behavior in any way

### DIFF
--- a/engine/NetworkProcessorNet.go
+++ b/engine/NetworkProcessorNet.go
@@ -91,6 +91,14 @@ func Peers(fnode *FactomNode) {
 		if fnode.State.GetHighestCompletedBlk() < fnode.State.GetTrueLeaderHeight()-35 {
 			// Discard all commits, reveals, and acks <= the highest ack height we have seen.
 			switch amsg.Type() {
+			case constants.COMMIT_CHAIN_MSG:
+				return true
+			case constants.REVEAL_ENTRY_MSG:
+				return true
+			case constants.COMMIT_ENTRY_MSG:
+				return true
+			case constants.EOM_MSG:
+				return true
 			case constants.MISSING_DATA:
 				if !fnode.State.DBFinished {
 					return true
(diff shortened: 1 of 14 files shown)


WhoSoup:
apologies, commit https://github.com/FactomProject/factomd/pull/647/commits/4306235a2ab450b97f5c2d0d1f5e6d4024c4aa20 looks very ugly in the diff due to the removal of the indenting. the only thing that should have changed was the removal of the `for i := 0; i < 1; i++ {` loop and closing brace

carryforward:
awesome, thank you.  This is still being tracked, but we are trying to get the next big release out before merging in these cleanup things.

keep em coming!

When this one gets pulled in we can get another branch to collect legibility things.

engine/NetworkProcessorNet.go
```diff
@@ -91,14 +91,6 @@ func Peers(fnode *FactomNode) {
 		if fnode.State.GetHighestCompletedBlk() < fnode.State.GetTrueLeaderHeight()-35 {
 			// Discard all commits, reveals, and acks <= the highest ack height we have seen.
 			switch amsg.Type() {
-			case constants.COMMIT_CHAIN_MSG:
```
factom-clay: Seems like this change makes the other message types return true when they would have fallen out and returned false.

WhoSoup: you are right. i have reversed these changes with https://github.com/FactomProject/factomd/pull/647/commits/a36264176241558509c57401e3f05b44e615009c

state/stateConsensus.go
```diff
@@ -255,11 +255,8 @@ func (s *State) Process() (progress bool) {
 	// is so we don't conflict with past version of the network if we have to reboot the network.
 	var Leader bool
 	var LeaderVMIndex int
-	if s.CurrentMinute > 9 {
```
Emyrk: I first thought this was needed as a minute ceiling, but the function `GetVirtualServers` does this, so doing it outside is redundant :+1:

state/validation.go
```diff
@@ -75,73 +75,70 @@ func (state *State) ValidatorLoop() {
 		// Look for pending messages, and get one if there is one.
 		var msg interfaces.IMsg
 
-		for i := 0; i < 1; i++ {
```
Emyrk: Validator loop went through refactor, so this is no longer relevant

carryforward:
ok, @WhoSoup There is a new branch to make legibility pulls against.  here it is:

https://github.com/FactomProject/factomd/tree/FD-1060_legibility_improvements_batch_2

PR #651, opened 2019-02-17
Add 2019-1 grants approved by standing parties to Fd 820 release candidate butter

nklomp:
These are all grants that have been approved by the standing parties during grant round 2019-1.
Addresses and amounts have been verified with recipients and reverification based on the code change has been requested to all recipients.

Please double check activation height of 180993. Activation should happen at 23-02-2019 00:00 UTC (not before)

### DIFF
--- a/CLA.txt
+++ b/CLA.txt
@@ -30,3 +30,4 @@ Name                                       Github Account                       
 Paul Snow                                  PaulSnow                                     Factom Inc.                                  Founder and Architect of Factom
 Brian Deery                                carryforward                                 Factom Inc.
 Samuel Vanderwaal                          samuelvanderwaal                             Canonical Ledgers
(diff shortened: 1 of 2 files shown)


carryforward:
awesome! Thank you @nklomp.  I'll do the doublecheck on this today.

carryforward:
ah, something for next time.  Grants go out on blocks that are divisible by 25 + 1
as an example:
152751 grant round 1
168576 grant round 2
169901 grant round 2

https://github.com/FactomProject/factomd/blob/FD-820_release_candidate_butter/state/grants.go#L481

This was done to prevent grants from colliding with ANO server rewards in the same block.  There is a 10 KiB size limit on fct transactions, including the coinbase that we want to stay far away from.  (this is why the genesis block was broken up into several transactions)

these were coded at 180993, so I will bump them back a couple hours to 181001 so they fall on a correct block number.

PR #655, opened 2019-02-20
removed extra call to now()

factom-clay:


carryforward:
lets pull this into Parchment instead of Kraft

PR #658, opened 2019-02-26
FD-845 Audit/Follower Brainswap - Compare height of heartbeat against state

ThomasMeier:
The context of `m` and `is` is a bit low for a beginner, but I believe this does a proper comparison to avoid a panic when an heartbeat is ahead of the node's DBHeight. If this is not on target, let me know the context of `m` and `is` so I can dive back in.

### DIFF
--- a/common/messages/heartbeat.go
+++ b/common/messages/heartbeat.go
@@ -326,7 +326,9 @@ func (m *Heartbeat) FollowerExecute(is interfaces.IState) {
 		if auditServer.GetChainID().IsSameAs(m.IdentityChainID) {
 			if m.IdentityChainID.IsSameAs(is.GetIdentityChainID()) {
 				if m.SecretNumber != is.GetSalt(m.Timestamp) {
-					panic("We have seen a heartbeat using our Identity that isn't ours")
+					if is.GetDBHeightComplete() >= m.DBHeight {
+						panic("We have seen a heartbeat using our Identity that isn't ours")
+					}
 				}
 			}
 			auditServer.SetOnline(true)


carryforward:
eek, I think this goes in the opposite direction that we are thinking of.  The panic is there for a reason.  I have asked clay to look into this.

I think clay might also have had a little more sophisticated checking in mind.

factom-clay:
I think your fix is not really what we want. First you should is.GetLLeaderHeight() to get the current height not is.GetDBHeightComplete() which tell you how caught up the second pass of syncing at boot is. At best it will tell you two block behind the currently being constructed block.  is.GetLLeaderHeight() tells you the current height being built.

Kind of what I was thinking was look at the last height where you changed ID. If the heartbeat is before that you don't care. If it's after you activated the current id but not in the future, then panic if it matches. If it's from the future and your next ID change is before then then it's ok. If its from the future and there is no scheduled ID change then hold on to it till you get to that height and revisit if it is panic worthy at that time.

Could be argued if it's from the future then it's probably ok and we can just ignore it because if it really is bad it will still be bad when we get to the future and we will see another and die then. The whole saving it and rechecking would be pretty complicated. 
state.SetIdentityChainID() is the place it's it actually set and s.AckChange is the next (or prev) height of a planned change.  State.CheckForIDChange() is called on block boundaries to check if the config file had changed and there is a now a request to change id's at some height.

simTest.TestAuditBrainSwap() test the case that causes an unintended panic. It does not pass with your change btw.  Really should have a test that test the intended panic too but I didn't write one when I was working on it. TestAuditBrainSwap is not called on CircleCI because it is a know fail. Once we integrate your fix we will add it to the circle run to make sure we don't reintroduce the bug.

Drop me an email clay@factom if you have questions.

factom-clay:
~/go/src/github.com/FactomProject/factomd/simTest(FD-845_brainswap_audit_panic)$ go test --run AuditBrainSwap

sanderPostma:
I found something in AuditBrainSwap_test that does not make sense to me:
Before the swap I see the following list:
`FNode0:5-1
FNode01:5-1
FNode02:5-1
FNode03:5-1
FNode04:5-1
FNode05:5-1
Found Leader   0 455b7b
Found Leader   1 1570f8
Found Leader   2 8da6ed
Found Audit     3 aeaac8
Found Follower 4 f0b7e3
Found Follower 5 d2bc4e
`
Then there is a brainswap between A3 and F5 which results in:
`Found Leader   0 455b7b
Found Leader   1 1570f8
Found Leader   2 8da6ed
Found Follower 3 d2bc4e
Found Follower 4 f0b7e3
Found Audit     5 aeaac8`
The audit server became 5 and the follower 3, all good.
But then it tests if node 4 is a leader and this fails:
`if !state4.Leader {
  t.Error("Node 4 did not become a leader, but " + curLeader)
  }`
As far as I can see this test will always fail.

factom-clay:
I agree the test should check 3 became a follower and 5 became an audit.  4 should be unaffected. Guess we never had the test get that far.

factom-clay:
GetDBHeightComplete() is not the right height to check against the message. Should be s.GetLLeaderHeight()

sanderPostma:
update:
I actually wrote a test today for the intended panic, only as far as I can see using the current test platform has no way to test for expected panics inside goroutines within goroutines. So instead of going into panic I fiddled a bit in my local copy so the heartbeat code sends a -1 to the node shutdown channel instead so only the offending node is shut down.
After I got this working another issue was exposed and is what would probably happen when running live nodes outside the sim: the innocent heartbeat messages turn into torpedoes assuring mutual destruction, not one audit node dies but both. Now I am left with a unit test failing the correct way:
`--- FAIL: TestAuditBrainDuplication/Run_Brain_Duplication_Sim/Wait_For_Identity_Duplication (67.14s)

            simulation.go:399: found 0 audit servers, expected 1`

factom-clay:
I don't think the "mutually assured destruction" is all that likely on a
real network. For it to happen that way on a live network both nodes have
to send their heartbeat before processing the others nodes heartbeat.  As
goods as networks are, the windows is relatively small for the two
heartbeats to pass in the night. The jitter between clock is small, but the
timeout that eventually causes the heartbeat to be created is processed
like any other message and and is subject to the large jitter of the
message processing path. The net effect is to make for a small window where
mutual destruction assured. Normally one node will see the other nodes
heartbeat before the time out that would create his own heartbeat and only
one node will panic. That said both nodes quitting is not bad for the
network. Just means one fewer audits is available till the ANO restart one
of the nodes to fill the position. In simulation, of course, the jitter is
small unless you turn on message delay (jitter) from simcontrol "Fxxxx"
where xxx is maximum delay in milliseconds.

The change to shutdown instead of panic is an interesting solution to the
inability to catch the panic. I'll have to consider it for a bit but it
seem like a good way to allow testing for that class of errors where the
intention is for the node to exit and lowers the chance of exiting with a
incomplete(corrupt) database.

Thanks,

Clay



Clay Douglass
Senior Architect
clay@factom.com | https://factom.com <http://factom.com/>

sanderPostma:
A panic state shuts the node down faster than a shutdown decreasing the window it can send out another heartbeat. The unit test confirms this and now only one audit node dies.
What I basically did now in ValidatorLoop is:

func (state *State) ValidatorLoop() {
	defer func() {
		if r := recover(); r != nil {
			fmt.Println("A panic state occurred in ValidatorLoop.", r)
			shutdown(state)
		}
	}()

This catches all panics and closes the database gracefully in the process.

factom-clay:
Works for me. If it'as all ready let me know and please coordinate with
Brian on a pull request. Then I'll code review the rest of the change.

Have a great week!

Clay



Clay Douglass
Senior Architect
clay@factom.com | https://factom.com <http://factom.com/>

sanderPostma:
I have two more questions:
1. When a dup is found, shouldn't it always panic when ChangeAcksHeight <= DBHeightAtBoot ? If ChangeAcks is not at a sensible height means the node is not waiting for a scheduled brainswap and can just shut down.

2. When we do decide to skip the panic and after both audit nodes are caught up, can there be window where both nodes are actually on the same id until the heartbeat messages were sent, processed and one of the nodes goes offline? If so shouldn't there be a fail-safe to make sure that both nodes abstain from the election process until for a while until at least one has died and no duplicate-id heartbeats were detected for one or more blocks. I still have to study and test the election process in more detail. Am I right or is the election process bulletproof enough to never let it be confused by this situation?

factom-clay:
When a node is changing identity it happens on a block boundary before a
heartbeat for minute 0 is created. So if node A has ChangeAcksHeight = N
then at Block N, Minute 9 the heartbeat will have the old identity and at
block N+1 minute 0 the heartbeat (assuming the new identity is also an
audit identity) A's heartbeat will be the new identity. So if two nodes (A
and B) swap identities (fred and barney) at the same height (same
ChangeAckHeight) then they will never produce heartbeats for the same
height with the same identity. But if A is behind B then B may become fred
at N and see a heartbeat from A for N-1 that also says fred. That is not a
panic case. Similarly A may see a heartbeat from B for N (A is at N-1) that
says fred. That is ok too. The only panic case is A sees a heartbeat from B
at A's current height with the same ID or visa versa. It is valid for A and
B to swap identities every block if they want. DBHeightAtBoot isn't really
a consideration, it's just the current height LLeaderHeight and the Height
in the heartbeat.

Have a great day!
Clay


Clay Douglass
Senior Architect
clay@factom.com | https://factom.com <http://factom.com/>

factom-clay:
I meant to say:
... So if node A has ChangeAcksHeight = N then at Block N-1, Minute 9...



Clay Douglass
Senior Architect
clay@factom.com | https://factom.com <http://factom.com/>

Emyrk:
Ticket number merged and closed here: https://github.com/FactomProject/factomd/pull/770

PR #659, opened 2019-02-27
improving handling of malformed messages

WhoSoup:
I believe this prevents the problems described in FD-850, which turned out to be two-fold.

1) The acceptance of Acks's DBHeight has been moved underneath the check that filters out Acks in the distant future. The highest distance now accepted is capped at LeaderHeight + 5.  

2) The signature validation adds a sanity check to filter out messages that do not have a signature at all

### DIFF
--- a/common/messages/ack.go
+++ b/common/messages/ack.go
@@ -115,10 +115,6 @@ func (m *Ack) Validate(s interfaces.IState) int {
 		return -1
 	}
 
-	if s.GetHighestAck() < m.DBHeight {
-		s.SetHighestAck(m.DBHeight) // assume the ack isn't lying. this will make us start requesting DBState blocks...
-	}
-
 	delta := (int(m.DBHeight)-int(s.GetLeaderPL().GetDBHeight()))*10 + (int(m.Minute) - int(s.GetCurrentMinute()))
 
 	if delta > 50 {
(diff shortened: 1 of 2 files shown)


factom-clay:
The changing of the setting of HighestAck to never be more than 5 block in the future negatively impact boot time by preventing us from asking for all the blocks we are missing. It is understood, though perhaps not well commented :-) that HighestAck is venerable to attacks that lie about the height other servers have. Probably in the syncing code we should have some policy to drag the HighestAck back to current height when we have timed out trying to get DBStates that are in our future. The down side to having a highestAck in the future is the  sync mechanism will continually periodically try to get the missing blocks. Please undo this change and request a new review.

Change 2 looks possibly valid. Looking at the code an unsigned  message looks like it would cause a panic without this change. But, I'd expect  it would fail to un-martial the message and thus a message with no signature could never exist ... unless created on the node itself, then it's a bug. It would be interesting to see a test that created an unsigned unsigned messages and sent it to a node to see what happens. I think verifying that none of the martial/un-martial code allow messages with missing signatures would be a good task too.

factom-clay:
If you revert change 1 and resubmit I'll approve the pull request.

WhoSoup:
> The changing of the setting of HighestAck to never be more than 5 block in the future negatively impact boot time by preventing us from asking for all the blocks we are missing. It is understood, though perhaps not well commented
I did not consider that and I am resubmitting a change that maintains this functionality but also still fixes the exploit mentioned in FD-850. Please review the updated code and let me know if this is acceptable.

The purpose of this new code is to cap the differential in highest ack that the system will start requesting. I picked 200 because that is also value used to throttle the actual requests in Catch(): https://github.com/FactomProject/factomd/blob/4ff747f2e0949fde6035af140aedcf64d26abc30/state/dbstateCatchup.go#L100-L102. Since ACKs keep coming in, this will keep being updated. I have tested it locally and a relative cap of 200 will not stall the network.

> It would be interesting to see a test that created an unsigned unsigned messages and sent it to a node to see what happens.
I have given such code to matt york via discord that triggers both of the issues I'm hoping to fix in this

PR #676, opened 2019-03-18
Fd 859 use peers.json file

WhoSoup:
I may have gone a bit outside of the scope of the original request.

Significant changes in this pull request:
* Re-enabling loading of peers.json file
* Only use the github seed on boot if the peers.json has less than "NumberPeersToConnect" peers (32) (note: this is probably not necessary and i'd be happy to reverse it)
* When discovering from the seed, don't overwrite the old peer but update it
* Added a test file to simulate loading from seed and file

Other changes:
* Fixes to stop nilpointer crashes while loading peers.json
* Changing Discovery mutex to an rwmutex and changing it from global to discovery-instance specific
* Respect errors thrown during peer initialization (dns lookup) and don't keep those peers

Not changed:
* Factomd discovers from github seed every 4 hours
* Qualityscore (it only saves peers with a score > -200, and resets them to 0 on connect)

### DIFF
--- a/p2p/connectionManager_test.go
+++ b/p2p/connectionManager_test.go
@@ -202,7 +202,8 @@ func TestConnectionManagerGetRandomRegularSingle(t *testing.T) {
 }
 
 func newPeer(address string, port string, peerType uint8) *Peer {
-	return new(Peer).Init(address, port, 100, peerType, 0)
+	p, _ := new(Peer).Init(address, port, 100, peerType, 0)
+	return p
 }
 
 func newIncomingConnection(peer *Peer) *Connection {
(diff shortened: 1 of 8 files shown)


WhoSoup:
Looks like it was saving to files with the key `ip:port` but when gathering from the seed and sharing peers with others, it just used `ip` without port. In commit https://github.com/FactomProject/factomd/pull/676/commits/d1c2e58a93eba33a3d9c8bbc335bb304d420f730 I fixed it so it's consistently using the currently expected format of `ip`, but I would like to change it across the board to use `ip:port`.

Any objections to that?

WhoSoup:
closing this pull request since the full implementation of what was actually requested requires a lot of additional work and it's likely this one won't be finished for quite a while

Issue #675, opened 2019-03-17
FD-925 Pprof is exposed by default via control panel port


WhoSoup:
By default, pprof is exposed regardless of the "`exposeprofiler`" setting if the control panel is exposed. The control panel and pprof share the same default mux and you can access `/debug/pprof` via port :8090. 

Steps to recreate:
1. Start factomd with default settings (exposeprofiler disabled)
2. Visit http://localhost:8090/debug/pprof/
3. (optional) Access remotely http://`address`:8090/debug/pprof/

Demonstration:
This is an unsecured control panel of a node in the [testnet seed file](https://raw.githubusercontent.com/FactomProject/communitytestnet/master/seeds/testnetseeds.txt): http://206.189.7.6:8090/. You can access pprof at http://206.189.7.6:8090/debug/pprof/ even though http://206.189.7.6:6060/debug/pprof/ is disabled.

Solution:
Create separate muxes for control panel and pprof, for a possible implementation, see: https://github.com/WhoSoup/factomd/commit/afceaf8cfabfbaa607b3382cf08b79d7f4939f50

carryforward:
This is a good change.  This has kind of been on the backburner as a thing to fix.  Thank you for doing this.  

This separates out the control panel mux, which is needed.  There are also the proflier, prometheus, and the api (8088 v1 and v2) ports which also might still share a mux.  Can we isolate those as well?

We need to be careful that we also don't break the https handling that the control panel and the API provide.  Can you verify that the https control panel still works with this update @WhoSoup ?

Can you reformulate this as a pull request to this branch? FD-925_use_different_web_service_muxes

PR #685, opened 2019-04-04
FD-925_use_different_web_service_muxes

WhoSoup:
Continuing from: https://github.com/FactomProject/factomd/issues/675

I have also added the separate mux for prometheus. The API already has a separate one, since it's using the [factomproject/web](https://github.com/FactomProject/web) server. 

HTTPs still works for both the control panel and the api, and pprof/prometheus are not accessible via port 8090 anymore, likewise for pprof and prometheus being able to access each other or the control panel. The API has always been separate and was not susceptible to this in the first place.


Emyrk:
I like this change :+1:

PR #687, opened 2019-04-09
Fd 926 sim test cleanup

stackdump:
This code looks good to me - though b/c I created the PR I can't approve.

EDIT: except for failing tests... :( if we'd like I continue to resolve in #FD-908 - which I'm still tweaking


stackdump:
rolled changes into this PR https://github.com/FactomProject/factomd/pull/756

PR #690, opened 2019-04-09
Fd 858 close tcp connection

factom-clay:
Clay reviewed this and had a few updates made. It looks good to me, though perhaps overly sophisticated on the peer matching. When we have a DevTestNet we should revisit the EOF behavior on receive. For now this seems better than what we had.


stackdump:
this may perhaps still be relevant

PaulBernier:
p2p has been replaced by p2p2 module in Wax release. Closing.

PR #692, opened 2019-04-09
Fd 827 fast cc cr

factom-clay:


factom-clay:
rebased and a new PR created.

PR #695, opened 2019-04-10
Fd 908 simtest cleanup whitelist

stackdump:
- Clean up Tests on top of what Clay's already done.

- Made a change to clean up output in circle - so we can see list of tests + runtimes

- Removed t.Run() - seemed to try to default to run in parallel on latest Goland

- Adjusted all tests to be named consistenly according to filename in ./simTest ./peerTest

- Altered default simulation behavior to use isolated factomHome dirs nested under $(PWD).sim/

- Added ./engine/debug/whitelist.txt as a means to specifiy which sim tests under ./engine - should be run on circle - NOTE: this could be adapted to drive the whole test suite on circle if needed - target from Demetri was to try to have a selection of tests that runs in 2 min


stackdump:
this PR is replaced by https://github.com/FactomProject/factomd/pull/733

PR #696, opened 2019-04-12
import test from old work

stackdump:
Add older test to Parchment

### DIFF
--- a/simTest/EntriesBeforeChain_test.go
+++ b/simTest/EntriesBeforeChain_test.go
@@ -29,8 +29,7 @@ func TestEntriesBeforeChain(t *testing.T) {
 
 	numEntries := 9 // set the total number of entries to add
 
-	params := map[string]string{"--debuglog": ""}
-	state0 := SetupSim("LLAAFF", params, 10, 0, 0, t)
+	state0 := SetupSim("LLAAFF", nil, 10, 0, 0, t)
 
 	var entries []interfaces.IMsg
 	var oneFct uint64 = factom.FactoidToFactoshi("1")
--- a/simTest/HoldingRebound_test.go
+++ b/simTest/HoldingRebound_test.go
@@ -26,9 +26,7 @@ func TestHoldingRebound(t *testing.T) {
 
 	println(a.String())
 
-	params := map[string]string{"--debuglog": "."}
-
-	state0 := SetupSim("L", params, 12, 0, 0, t)
+	state0 := SetupSim("L", nil, 12, 0, 0, t)
 
 	e := factom.Entry{
 		ChainID: id,
--- a/simTest/EntryBatch_test.go
+++ b/simTest/EntryBatch_test.go
@@ -31,8 +31,7 @@ func TestEntryBatch(t *testing.T) {
 
 	println(a.String())
 
-	params := map[string]string{"--debuglog": ""}
-	state0 := SetupSim("LLAAFF", params, 10, 0, 0, t)
+	state0 := SetupSim("LLAAFF", nil, 10, 0, 0, t)
 
 	var entries []interfaces.IMsg
 	var oneFct uint64 = factom.FactoidToFactoshi("1")
--- a/simTest/PermFCTBalancesAfterMin9Election_test.go
+++ b/simTest/PermFCTBalancesAfterMin9Election_test.go
@@ -1,0 +1,70 @@
+package simtest
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	. "github.com/FactomProject/factomd/engine"
+	. "github.com/FactomProject/factomd/testHelper"
+	"github.com/stretchr/testify/assert"
+)
+
+var bankSecret string = "Fs3E9gV6DXsYzf7Fqx1fVBQPQXV695eP3k5XbmHEZVRLkMdD9qCK"
+var depositAddresses []string
+
+// generate addresses & private keys
+func createDepositAddresses() {
+	for i := 0; i < 1; i++ {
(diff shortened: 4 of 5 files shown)


stackdump:
Will rebase against latest build if this test is still desirable

.circleci/config.yml
```diff
@@ -78,8 +78,9 @@ jobs:
             export PATH="/tmp/go/bin:$PATH"
             export GOPATH=/tmp/go
             cd /tmp/go/src/github.com/FactomProject/factomd/
-            go test $(glide nv | grep -v Utilities | grep -v longTest | grep -v peerTest | grep -v simTest) -vet=off -v -cover -coverprofile=coverage.out
+            go test $(glide nv | grep -v Utilities | grep -v longTest | grep -v peerTest | grep -v simTest | grep '...') -vet=off -v -cover -coverprofile=coverage.out
```
factom-clay: grep '...' matches any three characters. Probaly you want grep -E "[\.]{3}" which matches three . characters or ... '\.\.\.' though you may have to double \ that to make it work in a script.

simTest/AuditBrainSwap_test.go
```diff
@@ -18,80 +17,36 @@ This test is useful for verifying that Leaders can swap without rebooting
...
+	for i := 0; i < 6; i++ { // build config files for the test
+		WriteConfigFile(i, i, "", t) // just write the minimal config
+	}
+
+	params := map[string]string{"--factomhome": globals.Params.FactomHome}
+	state0 := SetupSim("LLLAFF", params, 15, 0, 0, t)
+	state5 := engine.GetFnodes()[5].State // Get node 5
+	_ = state5
+
+	t.Log("Disabled test while known bug exists FD-845")
```
factom-clay: Is FD-845 fixed in Xuan? Should this be re enabled?

simTest/EntriesBeforeChain_test.go
```diff
@@ -29,89 +26,60 @@ func TestCreatEntriesBeforeChain(t *testing.T) {
...
-		println(b.String())
-		println(a.String())
-	})
+	println(b.String()) // print factoid & ec addresses
+	println(a.String())
 
-	// KLUDGE: using "LAF" causes timeout on CI
-	t.Run("Run sim to create entries", func(t *testing.T) {
-		state0 := SetupSim("L", map[string]string{"--debuglog": ""}, 200, 0, 0, t)
+	params := map[string]string{"--debuglog": ""}
```
factom-clay: Isn't no log the default? Can we just pass nil here?

simTest/EntriesBeforeChain_test.go
```diff
@@ -29,89 +26,60 @@ func TestCreatEntriesBeforeChain(t *testing.T) {
...
+	state0 := SetupSim("L", params, 8, 0, 0, t)
```
factom-clay: Probably should have LF so this is testes as both leader and follower

simTest/HoldingRebound_test.go
```diff
@@ -26,66 +26,52 @@ func TestSendingCommitAndReveal(t *testing.T) {
...
-		ticker := WatchMessageLists()
+	params := map[string]string{"--debuglog": ""}
 
-		if dropRate > 0 {
-			state0.LogPrintf(logName, "DROP_RATE:%v", dropRate)
-			RunCmd(fmt.Sprintf("S%v", dropRate))
-		}
+	// REVIEW: changing simulation to LAF doesn't always pass cleanly on circle
+	// this may just mean we shouldn't check for empty holding
+	state0 := SetupSim("L", params, 9, 0, 0, t)
```
factom-clay: Probably should be LF so we test leader and follower behavior.

test.sh
```diff
@@ -2,25 +2,26 @@
...
 function runTests() {
-
-
   if [[ "${CI}x" ==  "x" ]] ; then
+    # run locally
     TESTS=$({ \
-      glide nv | grep -v Utilities | grep -v longTest | grep -v peerTest | grep -v simTest; \
-      #ls simTest/*_test.go;\
-      #ls peerTest/*_test.go;\
+      glide nv | grep -v Utilities | grep -v longTest | grep -v peerTest | grep -v simTest | grep -v elections | grep -v activations | grep -v netTest | grep "..." ; \
```
factom-clay: Why exclude elections tests? again grep "..." does not match just ...

.circleci/config.yml
```diff
@@ -78,8 +78,9 @@ jobs:
...
             goveralls -coverprofile=coverage.out -service=circle-ci -repotoken=$COVERALLS_TOKEN
+            exit 0 # coverage should never fail
```
stackdump: This ended up not working... there is still a Ticket in the backlog to try to address this in some way

.circleci/config.yml
stackdump: I fixed this in latest version - going to just export this single test rather than getting this branch rebased

simTest/AuditBrainSwap_test.go
stackdump: I'll check this in the latest rollup

simTest/EntriesBeforeChain_test.go
stackdump: I'll clean up necessary params

simTest/EntriesBeforeChain_test.go
stackdump: set to `LLAAFF` in parchment

test.sh
stackdump: I copied what I found there originally where elections were excluded

PR #697, opened 2019-04-14
Configurable p2p incoming and outgoing flags

WhoSoup:
This change raises the cap of max incoming connections to accommodate intermediate network growth, as well as providing config and command line parameters to change these numbers for future adjustments or customization for seed nodes.

Changes:
* Global variable `p2p.MaxNumberIncomingConnections` raised from 150 to 200
* Configuration file variables: `App.P2POutgoing` and `App.P2PIncoming`
* Command line flags: `-p2pOutgoing` and `-p2pIncoming`

Order of evaluation is (from lowest to highest): p2p package default variables, config file, command line

### DIFF
--- a/common/globals/globals.go
+++ b/common/globals/globals.go
@@ -35,6 +35,8 @@ type FactomParams struct {
 	RuntimeLog               bool
 	Exclusive                bool
 	ExclusiveIn              bool
+	P2PIncoming              int
+	P2POutgoing              int
 	Prefix                   string
 	Rotate                   bool
 	TimeOffset               int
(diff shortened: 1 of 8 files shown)


factom-clay:
Was changing the number of peer to 200 driven by some data? Can you send me the data? In prior testing the network was no better when using more peers above a number quite a bit less than we use now,  in the 50s as I recall, but its been a long time ago.

WhoSoup:
> Was changing the number of peer to 200 driven by some data? Can you send me the data? In prior testing the network was no better when using more peers above a number quite a bit less than we use now, in the 50s as I recall, but its been a long time ago.

Yes and no. The 200 value is arbitrary and simply raising the current cap of 150. The issue stems from the current design of the Seed/Backhaul nodes, which are the gateway for any new node booting up. After some network analysis and talks with Brian, the seed nodes have between 110 and 125 active connections. Should they reach the limit of 150 inbound connections, it would be impossible for a node to connect to the network the normal way. 

That's a fairly low threshold attack vector, and with the upcoming ANO round, we are going to see a bunch more nodes enter the system. I agree that regular nodes don't need this setting to be high, which prompted this PR to let individual nodes (such as the seed nodes) raise the cap individually. More detailed data is in the discord chat with Brian and you.

Issue #671, opened 2019-03-13
Determine via API if factomd is "ready"


carryforward:
OK, it looks like FD-869 as it was implemented didn't actually solve the problem that the OpenNode team actually wanted solving.  Some commentary about that is posted here: https://github.com/FactomProject/factomd/issues/664

Lets talk about what the problem that is trying to be solved.  Lets not post psudocode or API structure returns, but instead try to nail down the problem we are trying to solve.

It isn't an easy problem to define, as following along with consensus is always a matter of subjectivity (bitcoin boils this subjectivity down to # of confirmations).  The subjectivity is harder to define in factom.

@ilzheev @jcheroske @ThomasMeier

ilzheev:
Okay, let’s try. 

Main aim: we need to know, when factomd node is ready to be enabled back into Open Node pool. 

Requirements:
1) Latest entryblock fetched locally, so any read response (like EC balance or entry ack) returns up-to-date information
2) Write request (like Commit Entry or Chain), that sent to the factomd, will be processed by network immediately and block won’t be missed.

jcheroske:
Anton is right on target. I would only add that there seem to be two things that we need. 

The first is a simple flag that conveys what Anton just mentioned. The flag's name should be decoupled from implementation terminology, so that the implementation is free to evolve. If you want to add additional flags that are tied to implementation details, that would probably be useful too. 

I realize that there is no perfect way to determine if the API is ready. What we want is not perfection, but an expert consensus that that flag is the best that can reasonably be implemented. If the core devs don't create such a flag, then it falls on devs such as myself to try and figure out the readiness of the API, given various bits of information. Not a good situation.

Second, additional system information is needed from factomd so that it's health can be assessed externally. Just because factomd says the API is ready doesn't make it so. My previous pseudo-code post was an attempt to ask for all of that relevant information in one shot, instead of having to make multiple calls.

Finally, please don't relegate this to the `diagnostics` call, unless you're planning on making that call a first-class API method. We are asking for a method that will become part of the main API, not some addition to the debug API.

ilzheev:
Any ideas on this?

TorHogne:
Having better API calls to query the internal status of factomd would be really useful for many projects as well as generally for ANOs.

We should be able to determine the following status:
Ignore/Wait/Syncing/Ready/etc.. As well as if it is a fed or audit.

The above should preferably be implemented with streaming support via web socket as well. If so, it would be beneficial to add things like pending transactions.

[...]

PR #698, opened 2019-04-19
Fd 869 add status check to api

ThomasMeier:
This arises from a discussion with Anton, Jay, Matt, and I about a status endpoint. This arises from discussion about #671 in which the solution is to create a standalone, separate program to determine readiness from facts about the server. This PR is to propose adding a `GET /status` endpoint to the node with a variety of information the user can find for determining readiness. This helps satisfy some elements that the open api project and other projects want to use, in particular:

1. A `GET` route to return 200. For now this is good enough singular point for checking liveness.
2. Returns a collection of information to give insight to the user about readiness.

This is based on a sketch so let me know if this needs to go to a different branch or get rebased or moved to a different path, etc., the relevant code is under `wsapi/wsapi.go`

### DIFF
--- a/wsapi/wsapi.go
+++ b/wsapi/wsapi.go
@@ -805,6 +805,7 @@ func HandleHeights(ctx *web.Context) {
 
 type statusResponse struct {
 	Version                      string
+	NodeName                     string
 	BootTime                     int64
 	CurrentTime                  int64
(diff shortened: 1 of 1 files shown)


stackdump:
TODO: 

1. add  docs about what each flag means
2. add db syncing flag 
3. Ignore mode flag

Issue #683, opened 2019-04-03
FD-922 2nd pass flagging issue on boot


carryforward:
Under certain circumstances, these were the symptoms:

The node is booted and it catches up with the federated servers (it is at least very close to being caught up.  it is within a block of the federated servers.)  The issue was also seen on federated servers themselves on reboot, so it is probably not an issue related to trying to catch up.

To observe, look at the control panel after the node has caught up with the first pass.  It has been observed when a node has previously been booted and synced to be current immediately before starting.  The 2nd pass graph will  show that it is caught up with the highest block height.  The 2nd pass will alternate between being behind the 1st pass for 4-8 seconds, then it will show that it advanced to 100% for 10 seconds or so until it reverts back to being behind.

It was not observed closely enough to see when the behavior stops.

This was not verified using the `factom-cli get heights` call.  This would give a different view into the running of the system. It would indicate if it were a control panel issue or if it were deeper in the code.

these were the parameters when starting: `factomd -stderrlog=err.txt -stdoutlog=out.txt -checkheads=false -startdelay=300`

The bug was observed in codebases from 6.2.0 as well as 6.2.2, which are radically different from each other.  

This was the bug as it was seen by Alex from Factoshi.
https://cdn.discordapp.com/attachments/468172332792414209/562947140712923156/sync-bug-6.2.2.gif

It was also seen during the Feb 2019 network pause.  It was likely contributing to federated servers not being able to progress after being restarted.
https://docs.google.com/document/d/1GnhfvlBAg_2gLn6PvFyMmsxVGkl_6cHMJleZnSg_DS0/edit

I have attached the pprof profile from when it was happening. flagging_while_booting.pdf This was the pprof command: `go tool pprof http://localhost:6060/debug/pprof/profile?seconds=30`

I give it a 15% chance that the **leveldb(*tableCompactionBuilder)** was part of what was causing this issue.

The bugfix is currently being attempted by Sander at BIF.

carryforward:
[flagging_while_booting.pdf](https://github.com/FactomProject/factomd/files/3039999/flagging_while_booting.pdf)

carryforward:
Pull requests can go against this branch.  FD-922_2nd_pass_flagging_control_panel

afenrir:
I have created a copy of the blockchain and savestate from a node where this issue reliably occurs. You can download it here: https://factoshi.nyc3.digitaloceanspaces.com/m2.tar.gz

PR #701, opened 2019-04-23
FD-922: Fixed 2nd pass flagging (EntryDBHeightComplete)

sanderPostma:
This is my first commit so I'm writing down all the details.
During my debug session I noticed that EntryDBHeightComplete is set to s.EntryDBHeightComplete = scan - 1. The first time after one or more new blocks were saved to the DB this makes sense because GoSyncEntries still has some work to do. But at the end of the function s.EntryDBHeightComplete is set to the highest block.
The next iteration of the loop will set it back to scan-1 and after the sync code completes back again to the highest block. When running a custom net with a small database the code in between executes so fast that this behavior is difficult to observe, but with a production database the EntryDBHeightComplete will be at -1 for quite some time.
Please see the comments in the code review on the "files changed" tab..

### DIFF
--- a/state/entrySyncing.go
+++ b/state/entrySyncing.go
@@ -209,7 +209,7 @@ func (s *State) GoSyncEntries() {
 
 			if firstMissing < 0 {
 				if scan > 1 {
-					if s.EntryDBHeightComplete < scan-1 {
+					if s.EntryDBHeightComplete < scan - 1 {
 						s.EntryDBHeightComplete = scan - 1
 					}
 					s.LogPrintf("EntrySync", "Scan EntryDBHeightComplete = %d", s.EntryDBHeightComplete)
--- a/state/entrySyncing.go
+++ b/state/entrySyncing.go
@@ -182,11 +182,12 @@ func (s *State) GoSyncEntries() {
 
 	for {
 
+		highestSavedBlk := s.GetHighestSavedBlk()
 		ESMissing.Set(float64(len(missingMap)))
 		ESMissingQueue.Set(float64(len(s.MissingEntries)))
 		ESDBHTComplete.Set(float64(s.EntryDBHeightComplete))
 		ESFirstMissing.Set(float64(lastfirstmissing))
-		ESHighestMissing.Set(float64(s.GetHighestSavedBlk()))
+		ESHighestMissing.Set(float64(highestSavedBlk))
 
 		entryMissing = 0
 
@@ -204,11 +205,13 @@ func (s *State) GoSyncEntries() {
 		firstMissing = -1
 
 	dirblkSearch:
-		for scan := start; scan <= s.GetHighestSavedBlk(); scan++ {
+		for scan := start; scan <= highestSavedBlk; scan++ {
 
 			if firstMissing < 0 {
 				if scan > 1 {
-					s.EntryDBHeightComplete = scan - 1
+					if s.EntryDBHeightComplete < scan-1 {
+						s.EntryDBHeightComplete = scan - 1
+					}
 					s.LogPrintf("EntrySync", "Scan EntryDBHeightComplete = %d", s.EntryDBHeightComplete)
 					start = scan
 				}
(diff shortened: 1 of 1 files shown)


state/entrySyncing.go
```diff
@@ -182,11 +182,12 @@ func (s *State) GoSyncEntries() {
 
 	for {
 
+		highestSavedBlk := s.GetHighestSavedBlk()
```
sanderPostma: Besides that GetHighestSavedBlk is more work than retrieving a variable it doesn't seem like a good idea that this counter changes during the loop iteration.

state/entrySyncing.go
```diff
@@ -204,11 +205,13 @@ func (s *State) GoSyncEntries() {
 		firstMissing = -1
 
 	dirblkSearch:
-		for scan := start; scan <= s.GetHighestSavedBlk(); scan++ {
+		for scan := start; scan <= highestSavedBlk; scan++ {
 
 			if firstMissing < 0 {
 				if scan > 1 {
-					s.EntryDBHeightComplete = scan - 1
+					if s.EntryDBHeightComplete < scan-1 {
```
sanderPostma: Check if  scan -1 is actually greater than the current value of EntryDBHeightComplete to ensure it doesn't time-travel.

state/entrySyncing.go
```diff
@@ -304,10 +307,10 @@ func (s *State) GoSyncEntries() {
 				}
 			}
 
-			if s.EntryDBHeightComplete%1000 == 0 {
-				if firstMissing < 0 {
-					start = scan + 1 // If nothing is missing at scan, make sure we don't process at scan again.
-					//Only save EntryDBHeightComplete IF it's a multiple of 100 AND there are no missing entries
+			if firstMissing < 0 {
+				start = scan + 1 // If nothing is missing at scan, make sure we don't process at scan again.
```
sanderPostma: I think this line is supposed to be outside the modulo condition. When it's inside the condition, the dirblkSearch block is executing continuously with just a 10ms delay. When it's outside the condition it only goes in after additional blocks were saved.

state/entrySyncing.go
```diff
@@ -317,7 +320,7 @@ func (s *State) GoSyncEntries() {
 		}
 		lastfirstmissing = firstMissing
 		if firstMissing < 0 {
-			s.EntryDBHeightComplete = s.GetHighestSavedBlk()
+			s.EntryDBHeightComplete = highestSavedBlk
 			s.LogPrintf("EntrySync", "firstMissing EntryDBHeightComplete = %d", s.EntryDBHeightComplete)
```
sanderPostma: Is the intention to log this line on every iteration? (In the old scenario like every 10ms)
If not I think it should be
if firstMissing < 0 {
			if s.EntryBlockDBHeightComplete < highestSavedBlk {
				s.EntryDBHeightComplete = highestSavedBlk
				s.LogPrintf("EntrySync", "firstMissing EntryDBHeightComplete = %d", s.EntryDBHeightComplete)
			}
			time.Sleep(10 * time.Millisecond)
		}

state/entrySyncing.go
```diff
@@ -317,7 +320,7 @@ func (s *State) GoSyncEntries() {
 		}
 		lastfirstmissing = firstMissing
 		if firstMissing < 0 {
-			s.EntryDBHeightComplete = s.GetHighestSavedBlk()
+			s.EntryDBHeightComplete = highestSavedBlk
```
sanderPostma: Is the intention to log this line on every iteration? (In the old scenario like every 10ms)
If not I think this should be inside a if s.EntryBlockDBHeightComplete < highestSavedBlk {} condtion

state/entrySyncing.go
```diff
@@ -204,11 +205,13 @@ func (s *State) GoSyncEntries() {
 		firstMissing = -1
 
 	dirblkSearch:
-		for scan := start; scan <= s.GetHighestSavedBlk(); scan++ {
+		for scan := start; scan <= highestSavedBlk; scan++ {
 
 			if firstMissing < 0 {
 				if scan > 1 {
-					s.EntryDBHeightComplete = scan - 1
+					if s.EntryDBHeightComplete < scan - 1 {
```
sanderPostma: Check if scan -1 is actually greater than the current value of EntryDBHeightComplete to ensure it doesn't time-travel.

carryforward:
wow, nice, thank you.  Were you able to replicate the bug using the databases that @afenrir provided in https://github.com/FactomProject/factomd/issues/683 ?

sanderPostma:
Yes I used that database and the behavior was very persistent. Actually I will take another look to see if there is a secondary problem or at least find out why this specific database exposed the unexpected behavior in entrySyncing.

afenrir:
It is worth noting that the database in question came from a factomd instance running on my laptop. That instance is stopped and started very frequently as I open and close my laptop, so it is forever going offline then catching up. Other than the OS (OSX), it is likely the largest environmental difference this database was subjected to relative to its server-based cousins, so may be worth considering.

carryforward:
oh, and would you please add your github ID to the CLA so that this fix can be merged in @sanderPostma?

Can you make a pull request to this branch and this file:
https://github.com/FactomProject/factomd/blob/Contributor_License_Agreement/CLA.txt

carryforward:
awesome, thank you for the contribution.  Lets get this pulled into the release.

PR #705, opened 2019-04-26
Fd 977 holding cleanup

factom-clay:
Clean up holding accounting, make ACKs execute from holding.


Emyrk:
See JIRA about how this was resolved

PR #707, opened 2019-04-29
Fd 975 dual ack bug fix

factom-clay:


factom-clay:
Created a branch that didn't include 976 and 977

PR #711, opened 2019-04-30
Fd 971 sim ctrl dropping messages

factom-clay:


stackdump:
this seems to improve simulator performance using the test created on this branch https://github.com/FactomProject/factomd/tree/FD-957_induce_stall

PR #718, opened 2019-05-03
configurable p2p incoming and outgoing flags

factom-clay:


factom-clay:
Inadvertently merged to master. Undid that and merged to parchment instead.

factom-clay:
Visited with Brian about your concern about the guard servers becoming full and thus preventing new nodes from joining the network.  My preferred fix would be for nodes to send the peer list before disconnecting a rejected peer.

PR #720, opened 2019-05-05
Fd 869 add ignore to control panel batch api

stackdump:
extend control panel to query ignore status


stackdump:
@carryforward  tested w/ 
``` watch -n 1 'curl http://127.0.0.1:8090/factomdBatch?batch=myHeight,leaderHeight,completeHeight,ignoreDone | python -m json.tool' 
```

This is much more limited than the previous `/status` call I showed you - I think we can get buy w/ this for now / maybe revisit the idea of `/status` addition to the RPC in the future

Emyrk:
This works as expected, can be beneficial :+1:

PR #721, opened 2019-05-09
backport simtests

stackdump:
Backport sim tests slated to go into Parchment

also note - currently this only runs the 'new' simulations in the ./simTest folder b/c I left this in it
```var RanSimTest = true // KLUDGE disables all sim tests during a group unit test run```

These two failures both pass for me locally - for Leaderbrain swap - there is still a chance for leader to see it's own identity on the network 

For Add Fnode / the test steals the DB from another fnode to boot up a new follower (is more reliable on local machine )

```
FAIL: simTest/LeaderBrainSwap_test.go 
FAIL: simTest/AddFNode_test.go
```


stackdump:
though we used this branch to test - we decided not to merge back into bond

PR #727, opened 2019-05-11
Fd 1013 bad comments dead redundent code

PaulSnow:


stackdump:
I think much of this has been replaced or has made it into previous releases

PR #735, opened 2019-05-15
FD-968 release candidate bond

carryforward:


carryforward:
this pull request is into master, and also includes the 6.2.2 which was merged into Develop. 
we should use the other pull request: https://github.com/FactomProject/factomd/pull/736

PR #745, opened 2019-05-28
FD-845 brainswap audit panic

sanderPostma:
improved handling of heartbeat panic
fixed and refactored AuditBrainSwap unit test
improved shutdown/db close handling in the validatorloop.go
introduced sequential run-states to have fine grained control


state/validation.go
```diff
@@ -137,14 +141,35 @@ func (state *State) ValidatorLoop() {
...
+	default:
+		return false
 	}
 }
 
+func shutdown(state *State) {
+	state.RunState = runstate.Stopping
+	fmt.Println("Closing the Database on", state.GetFactomNodeName())
+	state.DB.Close()
+	state.StateSaverStruct.StopSaving() // Shouldn't this be done before closing the database?
```
sanderPostma: Shouldn't this be done before closing the database? Or is there a specific reason to do this after?

sanderPostma:
Note:
I'm runnung the unit test using go test -run TestAuditBrainSwap -v
Running it inside Goland inconsistently fails the test even when all reported back PASS

state/validation.go
Emyrk: The savestate doesn't pull it's data from the db. So it isn't actually dependent on the db

Emyrk:
Going to merge this into an FD branch for a rebase and final code review :+1:

PR #746, opened 2019-05-29
Runtimelog doesn't respect its own enabled setting

WhoSoup:
The command line flag -runtimeLog lets you enable/disable the runtimelog, which is the log of messages you can print via the console "m" command. It buffers 4 seconds worth of messages internally, after which the buffer is cleared.

At the moment, the function that buffers message doesn't respect the setting whether or not it's enabled. This fix will:
* Disable message buffering when the runtimeLog flag is turned off, saving a little bit of memory and cpu
* Print out an info message if someone turns on message logging in the console with the setting disabled

### DIFF
--- a/engine/MsgLogging.go
+++ b/engine/MsgLogging.go
@@ -53,6 +53,9 @@ func (m *MsgLog) Init(enable bool, nodecnt int) {
 }
 
 func (m *MsgLog) Add2(fnode *FactomNode, out bool, peer string, where string, valid bool, msg interfaces.IMsg) {
+	if !m.Enable {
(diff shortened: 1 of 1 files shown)


WhoSoup:
> By default, are we always going through the overhead of maintaining this msg list?

yes

Emyrk:
I will get back to this PR. Any reviewer please note the base branch is currently set to master.

Emyrk:
@WhoSoup did you get an FD number for this?

Emyrk:
Tracking with FD-1069

PR #753, opened 2019-06-12
added mutex locks for list adds and dels

michaelbeam:


Emyrk:
Adds/Dels are now protected :+1:

PR #756, opened 2019-06-13
Fd 828 dependant holding bond rebase

stackdump:
Ready for Merge into Parchment


testHelper/simulation.go
```diff
@@ -237,40 +256,52 @@ func TimeNow(s *state.State) {
...
+		fmt.Fprintf(os.Stdout, "Printing status from %s", s.FactomNodeName)
+		if !StatusEveryMinuteRunning {
+			StatusEveryMinuteRunning = true
+			go func() {
+				for {
+					statusMutes.Lock()
+					s = statusState
+					statusMutes.Unlock()
+					if s == nil {
+						break // exit the routine
```
stackdump: @Emyrk wondering if changes here will fix the issue you're seeing w/ sim test

Emyrk: Yes it will :+1: 

There is a similar fix in parchment. Either or will suffice :man_shrugging:

stackdump:
I changed the target for merge - I'll resolve the conflicts & update the branch again

common/messages/ack.go
```diff
@@ -135,7 +135,7 @@ func (m *Ack) Validate(s interfaces.IState) int {
 	}
 
 	if delta > 30 {
-		return 0 // put this in the holding and validate it later
```
stackdump: Does this threshold still make sense w/ new Dependent holding?

stackdump:
Seeing coverage task fail  ```coverage: 72.3% of statements
ok  	github.com/FactomProject/factomd/database/mapdb	11.193s	coverage: 72.3% of statements
FAIL	github.com/FactomProject/factomd/database/securedb [build failed]```

common/messages/ack.go
factom-clay: we need some limit. I have always thought that more than 11 minutes into the future was probably pointless but Paul wanted 30 so here we are.

common/messages/revealEntry.go
```diff
@@ -186,15 +191,20 @@ func (m *RevealEntryMsg) Validate(state interfaces.IState) int {
 		if eb == nil {
 			state.LogMessage("executeMsg", "Hold, no chain", m)
 			// No chain, we have to leave it be and maybe one will be made.
-			return 0
+			//old holding .., return 0
```
Emyrk: If there is no chain, we still accept the commit.
So for the dependent holding, will this reveal ever get out? 

Should it be dependent on the chain instead of the entry hash? (Would need to keep a list of entries depending on a chain then)

state/HoldingList.go
```diff
@@ -0,0 +1,212 @@
...
+
+	/*
+		REVIEW:
+		Maybe we should treat the message stream as a votes on the "highest known block" where known servers trump unknown servers who disagree?
+
+		Consider setting HKB and HAB when we complete minute 1 of a block to the current leader height.
+		That at least would make us recover from a spoofed ack attack.
+	*/
+
+	switch msg.Type() {
```
Emyrk: How do Entry commits/reveals get expired?

state/factoidstate.go
```diff
@@ -324,14 +325,25 @@ func (fs *FactoidState) UpdateTransaction(rt bool, trans interfaces.ITransaction
...
 		fs.State.LogPrintf("entrycredits", "At %d process %s", fs.DBHeight, trans.String())
 	}
 	for _, ecOut := range trans.GetECOutputs() {
 		ecbal := int64(ecOut.GetAmount()) / int64(fs.State.FactoshisPerEC)
-		fs.State.PutE(rt, ecOut.GetAddress().Fixed(), fs.State.GetE(rt, ecOut.GetAddress().Fixed())+ecbal)
+		adr := ecOut.GetAddress().Fixed()
+		fs.State.PutE(rt, adr, fs.State.GetE(rt, adr)+ecbal) // Add EC's from FCT
+
+		// execute any messages that were waiting on this EC address
+		if rt == true {
```
Emyrk: Is there a reason why we check if rt is true here, but not for the fct outputs?

common/messages/revealEntry.go
Emyrk: Nvm, it does do that

state/HoldingList.go
```diff
@@ -0,0 +1,212 @@
...
+	case constants.DIRECTORY_BLOCK_SIGNATURE_MSG:
+		if msg.(*messages.DirectoryBlockSignature).DBHeight < l.s.GetHighestKnownBlock()-1 {
+			res = true
+		}
+	default:
+		l.s.LogMessage("newHolding", "SKIP_DBHT_REVIEW", msg)
+	}
+
+	if msg.GetTimestamp().GetTime().UnixNano() < l.s.GetFilterTimeNano() {
+		res = true
```
stackdump: this will expire messages that don't have db HT

state/HoldingList.go
Emyrk: By timestamp below. Got it :+1:

stackdump: they expire by timeout via timestamp rather than height

.circleci/config.yml
```diff
@@ -78,8 +78,9 @@ jobs:
             export PATH="/tmp/go/bin:$PATH"
             export GOPATH=/tmp/go
             cd /tmp/go/src/github.com/FactomProject/factomd/
-            go test $(glide nv | grep -v Utilities | grep -v longTest | grep -v peerTest | grep -v simTest) -vet=off -v -cover -coverprofile=coverage.out
+            go test $(glide nv | grep -v Utilities | grep -v longTest | grep -v peerTest | grep -v simTest | grep '...') -vet=off -v -cover -coverprofile=coverage.out
```
Emyrk: Is the `grep ...` supposed to grep the regex of any 3 character sequence? Or grep for `grep "\.\.\."`

stackdump: it looks for three literal '...' - one of the packages returned from `glide -nv` is "."

state/factoidstate.go
stackdump: we spoke about this - commenting for posterity
w/ the condition in place - only changes to temporary balance would result in triggering an execution of held messages
we ended up removing this so execution will be attempted for any balance change

this is to account for an apparent edge case where the balance changes because the node loaded a Db state rather then processing messages

Emyrk:
Pending final QA approval

PR #757, opened 2019-06-14
FD-809: Inflation limit

sanderPostma:
Initial proposal.
Please check the comments inside the "Files changed" and let me know it it matches what you expected. I've also considered keeping track of the inflation over a longer history of blocks. But this seems like a task for a monitoring /alert tool, however we could auto-cancel all payouts when the inflation rate over the last x number of blocks goes over a certain limit .

### DIFF
--- a/common/constants/constants.go
+++ b/common/constants/constants.go
@@ -278,6 +278,12 @@ var (
 	// 		:: Default = 6.4*1e8
 	COINBASE_PAYOUT_AMOUNT = uint64(6.4 * 1e8)
 
+	// Limited number of 100 fed servers in the network at 0% efficiency
+	COINBASE_PAYOUT_SERVER_LIMIT = uint64(100) * COINBASE_PAYOUT_AMOUNT
+
+	// Total grant limit for one block, 9000000000000 is the highest ever recorded
+	COINBASE_PAYOUT_GRANT_LIMIT = uint64(9000000000000)
+
 	// The height at which coinbase transactions will activate.
 	//	 This is useful for updating without needing to take
 	// 	 down the network and giving an update period.
--- a/state/dbStateManager.go
+++ b/state/dbStateManager.go
@@ -844,7 +844,7 @@ func (list *DBStateList) FixupLinks(p *DBState, d *DBState) (progress bool) {
 				continue
 			}
 			amt := primitives.CalculateCoinbasePayout(ia.Efficiency)
-			if amt == 0 {
+			if amt == 0 || amt > constants.COINBASE_PAYOUT_SERVER_LIMIT {
 				continue
 			}
 
--- a/state/factoidstate.go
+++ b/state/factoidstate.go
@@ -436,6 +436,18 @@ func (fs *FactoidState) GetCoinbaseTransaction(dbheight uint32, ftime interfaces
 				m[v] = struct{}{}
 			}
 
+			totalGrantAmount := uint64(0)
+			for i, o := range desc.Outputs {
+				if _, ok := m[uint32(i)]; !ok {
+					totalGrantAmount += o.GetAmount()
+				}
+			}
+
+			// Auto-cancel when the total amount exceeds the grant limit
+			if totalGrantAmount > constants.COINBASE_PAYOUT_GRANT_LIMIT {
+				return coinbase
+			}
+
 			for i, o := range desc.Outputs {
 				// Only elements not in map are ok
 				if _, ok := m[uint32(i)]; !ok {
--- a/state/grants.go
+++ b/state/grants.go
@@ -314,8 +314,11 @@ func CheckGrants() {
 
 //return a (possibly empty) of coinbase payouts to be scheduled at this height
 func GetGrantPayoutsFor(currentDBHeight uint32) []interfaces.ITransAddress {
-
 	outputs := make([]interfaces.ITransAddress, 0)
+
+	// Get combined limit of all hardcoded grants
+	totalGrantAmount := uint64(0)
+
 	// this is only but temporary, once the hard coded grants are payed this code will go away
 	// I can't modify the grant list because in simulation it is shared across nodes so for now I just
 	// scan the whole list once every 25 blocks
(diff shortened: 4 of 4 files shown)


common/constants/constants.go
```diff
@@ -278,6 +278,12 @@ var (
 	// 		:: Default = 6.4*1e8
 	COINBASE_PAYOUT_AMOUNT = uint64(6.4 * 1e8)
 
+	// Limited number of 100 fed servers in the network at 0% efficiency
+	COINBASE_PAYOUT_SERVER_LIMIT = uint64(100) * COINBASE_PAYOUT_AMOUNT
```
sanderPostma: The correct ceiling for the payouts have to be determined as there is almost no coming back from this once it's rolled out.

state/dbStateManager.go
```diff
@@ -844,7 +844,7 @@ func (list *DBStateList) FixupLinks(p *DBState, d *DBState) (progress bool) {
 				continue
 			}
 			amt := primitives.CalculateCoinbasePayout(ia.Efficiency)
-			if amt == 0 {
+			if amt == 0 || amt > constants.COINBASE_PAYOUT_SERVER_LIMIT {
```
sanderPostma: When the fed server payout that occurs every 25 blocks exceeds the limit, don't load it onto the admin block.

state/factoidstate.go
```diff
@@ -436,6 +436,18 @@ func (fs *FactoidState) GetCoinbaseTransaction(dbheight uint32, ftime interfaces
...
+			totalGrantAmount := uint64(0)
+			for i, o := range desc.Outputs {
+				if _, ok := m[uint32(i)]; !ok {
+					totalGrantAmount += o.GetAmount()
+				}
+			}
+
+			// Auto-cancel when the total amount exceeds the grant limit
+			if totalGrantAmount > constants.COINBASE_PAYOUT_GRANT_LIMIT {
+				return coinbase
```
sanderPostma: Ignore grant payouts altogether when total amount exceeds the grant limit

state/grants.go
```diff
@@ -324,6 +327,14 @@ func GetGrantPayoutsFor(currentDBHeight uint32) []interfaces.ITransAddress {
 	// there is no need for activation height because the grants have inherent activation heights per grant
 	for _, g := range GetHardCodedGrants() { // check every hardcoded grant
 		if g.DBh == currentDBHeight { // if it's ready {...
+
+			// Calculate if the total grant amount does not exceed the total grant limit
+			totalGrantAmount += g.Amount
+			if totalGrantAmount > constants.COINBASE_PAYOUT_GRANT_LIMIT {
+				outputs = make([]interfaces.ITransAddress, 0)
```
sanderPostma: Ignore grant payouts altogether when total amount exceeds the grant limit

sanderPostma:
Please check the comments inside the "Files changed" and let me know it it matches what you expected. I've also considered keeping track of the inflation over a longer history of blocks. But this seems like a task for a monitoring /alert tool, however we could auto-cancel all payouts when the inflation rate over the last x number of blocks goes over a certain limit .

state/dbStateManager.go
Emyrk: That `amt` is coming from calculating the coinbase payout on a per authority basis. Not the sum of all the payouts.

Your comment on `COINBASE_PAYOUT_SERVER_LIMIT` says that it is the payout limit of 100 fed nodes, and this conditional is checking the 100 fed node limit against a single node payout. 

Either you'd want to change the conditional to be checking the limit of a single node, or summation all the payouts, and then do a check.

state/factoidstate.go
Emyrk: Is the limit supposed to be set on a grant round by grant round basis?

So when we hardcode in the next round of grants, would we also update `COINBASE_PAYOUT_GRANT_LIMIT` if there is a new limit?

sanderPostma: As long as these limits are internal I don't think there is coming back from it after these versions have been deployed. Followers running an older version which do not know about a higher COINBASE_PAYOUT_GRANT_LIMIT will not save these blocks to the database until they are upgraded.
Maybe it's a better idea is to go back to the drawing board and allow the limits to be set in an adminblock entry just like ANO efficiency values are altered. But I can't determine how big the impact will be for nodes running older versions and if a change like that is desirable as it's effectively a protocol change.

Emyrk: Good point. I think adding admin block changes is a bit overkill, we do have the ability to cancel assuming we catch it in time.

I think some discussion around this would be beneficial to determine what the limit is.
Best case scenario is we calculate the amount in the grant pool, and use that as the limit, rather than a hard coded ceiling.

carryforward:
These limits on inflation are from the governance document:
https://docs.google.com/document/d/1RVaVR7lvfGgOBMG-7oca9TtpnR7qaEfr6XJVaZJwd3M/edit#heading=h.v1q4mc67184j

```
Token Supply
The Token supply will grow through a fixed set of awards that amounts to a 10% inflation of the Factoid supply in the first year, without considering usage of the protocol (which burns Factoids).
The protocol support budget is fixed at about 72933.12 tokens per month, where a month is defined as 4383 blocks. (365.25*24*60 / 10 / 12)
In many blockchains, Proof of Work accounts for much of the rewards issued for the security of the blockchain.  In other words, most of the resources are expended on energy costs rather than development, maintenance, and infrastructure.  As the protocol uses anchoring, resources can be expended on a sort of “Proof of Development,” extending and developing the protocol and the ecosystem around the protocol.  This is the motivation around the Grant Pool design.
Token Rewards
The Protocol issues tokens for each server in the Authority Set.  The issuance would be fixed at .256 Factoids per Authority Server per block.  If one assumes a year to be 365.25 days, 10 minute blocks, and 65 servers in the authority set, then 72933.12 tokens will be created per month, and 874,598.4 tokens per year. This is roughly a 10 percent inflation rate for the first year of token distribution. 
Grant Pool Allocation
```

carryforward:
this also gives a visual of how the grant pool dynamics work, where the grant pool builds up and then is paid out, but always less than the inflation limit.
https://www.factoshi.io/factoids

sanderPostma:
> The protocol support budget is fixed at about 72933.12 tokens per month, where a month is defined as 4383 blocks. 

=> So that's 16,64 per block.
At the time my my calculation attempt there were 57318 blocks between the coinbase activation and now. So the total budget would be 953771,52
The payouts until then (none were canceled) amount to 898021
There should be 55750,52 left, only the Factoshi site says it's around 26000 today, that's roughly 20000 less, so I think I'm missing something. Is efficiency taken into account there?

Emyrk:
I had thought factoshi.io did account for efficiency. But I have not confirmed that

stackdump:
We need to revisit this 
We'd like to have tighter bounds - 
you should be able to calculate what the grant pool is and prevent it from going sub-zero

PR #758, opened 2019-06-14
FD-621 : Correct BalanceFinder utility to compile

Emyrk:
Also refactored CorrectChainHeads to use same tool code
as BalanceFinder. Refactored to match new function definitions in
/factom


carryforward:
this looks good.

PR #759, opened 2019-06-14
Fd 1058 simtest shutdown

Emyrk:
Correct unit test sporadic failures


simTest/HoldingRebound_test.go
```diff
@@ -67,7 +67,7 @@ func TestSendingCommitAndReveal(t *testing.T) {
 			t.Run("Fund ChainCommit Address", func(t *testing.T) {
 				amt := uint64(11)
 				engine.FundECWallet(state0, b.FctPrivHash(), a.EcAddr(), amt*state0.GetFactoshisPerEC())
-				WaitForAnyDeposit(state0, a.EcPub())
+				WaitForEntry(state0, reveal.GetHash())
```
stackdump: good call - this is a better test

PR #761, opened 2019-06-17
FD-406 split receipt and anchor api

sambarnes:
Originally created back in fall 2018, and had to keep getting rebased onto newer branches as it sat in code review.

This latest iteration is off of Bond v6.3.1-rc1 (which is now actually 6.3.2)


sambarnes:
Before this gets merged, we also need to set the key that is going to be used for ethereum anchoring, if it's different from the one that is shown here for "m2":

https://github.com/FactomProject/factomd/blob/FD-406_ReceiptAndAnchorAPI_v6.3.1-rc1/database/databaseOverlay/anchorInfo.go#L16

sambarnes:
Added the following:

- command line flag `--reparseanchorchains` which, when set, will drop all previously validated and stored anchors from the database, and reparse both bitcoin and ethereum anchor chains according to the custom or default anchor record keys

- config file options `BitcoinAnchorRecordPublicKeys` and `EthereumAnchorRecordPublicKeys` which are expected to be public keys in hex

These config file variables are arrays. To specify multiple values in gcfg, you would:
```
BitcoinAnchorRecordPublicKeys = "1234567890abcdef..."
BitcoinAnchorRecordPublicKeys = "abcdef1234567890..."
```

If no customs are specified, the following configuration is used:
```
BitcoinAnchorRecordPublicKeys         = "0426a802617848d4d16d87830fc521f4d136bb2d0c352850919c2679f189613a" ; m1 key
BitcoinAnchorRecordPublicKeys         = "d569419348ed7056ec2ba54f0ecd9eea02648b260b26e0474f8c07fe9ac6bf83" ; m2 key, currently in use
EthereumAnchorRecordPublicKeys        = "a4a7905ab2226f267c6b44e1d5db2c97638b7bbba72fd1823d053ccff2892455"
```

wsapi/wsapiV2.go
```diff
@@ -736,24 +708,196 @@ func HandleV2RawData(state interfaces.IState, params interface{}) (interface{},
...
+			var allWindowKeyMRs []interfaces.IHash
+			for i := eth.DBHeightMin; i <= eth.DBHeightMax; i++ {
+				keyMR, err := dbo.FetchDBKeyMRByHeight(uint32(i))
+				if err != nil {
+					return nil, NewCustomInternalError(err)
+				}
+				allWindowKeyMRs = append(allWindowKeyMRs, keyMR)
+			}
+			eth.MerkleBranch = primitives.BuildMerkleBranchForHash(allWindowKeyMRs, directoryBlockKeyMR, true)
+			response.Ethereum = eth
```
Emyrk: This looks like a code duplicate of what is below. Can we refactor into a function so we don't have to modify 2 pieces of the same code if we need a change?

sambarnes: Thanks for the reminder there! Just updated it if ya wanted to take a look

Emyrk:
rebased here: https://github.com/FactomProject/factomd/pull/781

PR #763, opened 2019-06-17
Fd 1037 additional brainswap tests

sanderPostma:
I still had a branch with an additional test to study the behavior in case of a bad node upgrade procedure where and audit server is duplicated / the initial one is left online or accidentally comes back online.
I experimented with heartBeat.go & syncMessage.go to a seemingly more desirable outcome. I'm unsure  if my changes make sense but I'll just submit them foe review. Merge the parts that you think are useful.

Only this was done on top of FD-845 and contains the same fixes because it's branched off another version.


common/messages/heartbeat.go
```diff
@@ -326,7 +327,19 @@ func (m *Heartbeat) FollowerExecute(is interfaces.IState) {
 		if auditServer.GetChainID().IsSameAs(m.IdentityChainID) {
 			if m.IdentityChainID.IsSameAs(is.GetIdentityChainID()) {
 				if m.SecretNumber != is.GetSalt(m.Timestamp) {
-					panic("We have seen a heartbeat using our Identity that isn't ours")
+					lLeaderHeight := is.GetLLeaderHeight()
+					ackChangeValue := is.GetAckChangeValue()
+					ackChangeScheduled := ackChangeValue > is.GetDBHeightAtBoot() && lLeaderHeight <= ackChangeValue
```
sanderPostma: Loosen the panic condition so the node with the highest chkChange value will not panic right away but waits until this happens again after the block when the brainswap is done.

common/messages/heartbeat.go
```diff
@@ -326,7 +327,19 @@ func (m *Heartbeat) FollowerExecute(is interfaces.IState) {
...
+					if m.DBHeight == lLeaderHeight && (!ackChangeScheduled || is.GetDuplicateIdDetectedAtHeight() > ackChangeValue) {
+						var b strings.Builder
+						fmt.Fprintf(&b, "We have seen a heartbeat using our Identity that isn't ours.\n")
+						fmt.Fprintf(&b, "    Node: %s\n", is.GetFactomNodeName())
+						fmt.Fprintf(&b, "    LLeaderHeight: %d\n", lLeaderHeight)
+						fmt.Fprintf(&b, "    Message dbHeight: %d\n", m.DBHeight)
+						panic(b.String())
```
Emyrk: After thinking about it more, we have a feature coming in parchment that allows us to hold a message until a given height. We are working towards a refactor that removes state in areas where it is not needed, so I'd prefer to have this message be held until the height is reached. This removes our height checks here.

Would this also solve the `ackChangeScheduled`? I don't I entirely understand what this logic is aiming for.

stackdump:
@clay @Emyrk  do we want this in a4? or do other changes to the system address this?

PR #765, opened 2019-06-17
Reintegrate the successful Bond merge back into Parchment

carryforward:


Emyrk:
FD-1053_bond_parchment_merge has been the head of parchment for a little while. Getting it back into the original feature branch makes tracking easier :+1:

PR #770, opened 2019-06-18
FD-845 brainswap audit panic rebased

Emyrk:
Continuation of https://github.com/FactomProject/factomd/pull/745


Emyrk:
FD-1037 PR has these commits + additional changes

https://github.com/FactomProject/factomd/pull/763/commits

Emyrk:
I realize now this is a subset of 1037. I was too quick to close it. I am going to reopen as the changing of IsRunning is a good change, and is more intuitive way of tracking the current state.

common/messages/heartbeat.go
```diff
@@ -326,7 +327,15 @@ func (m *Heartbeat) FollowerExecute(is interfaces.IState) {
 		if auditServer.GetChainID().IsSameAs(m.IdentityChainID) {
 			if m.IdentityChainID.IsSameAs(is.GetIdentityChainID()) {
 				if m.SecretNumber != is.GetSalt(m.Timestamp) {
-					panic("We have seen a heartbeat using our Identity that isn't ours")
+					lLeaderHeight := is.GetLLeaderHeight()
+					if m.DBHeight == lLeaderHeight {
```
Emyrk: I think this change is valid, however, dependent holding is finalized and will be in parchment. With that, we can hold this message until we hit the height it is for. This would remove this state logic from this msg execute, which is the direction we are headed with the refactor. When the dependent holding is apart of the new parchment head, I say we move this.

Emyrk: Until this is resolved in dependent holding, this check is valid :+1:

Emyrk:
`filepath.Join()` makes some file concatenation stuff way easier -> https://play.golang.org/p/SYgpcDqIe8e

PR #773, opened 2019-06-20
Fd 890 where logs go rebase

JoshuaBrigati:


common/messages/messageTrace.go
```diff
@@ -39,38 +39,49 @@ func checkFileName(name string) bool {
...
+		flag = TestRegex.Match([]byte(name))
+		enabled[name] = flag
+	}
+	return flag
+}
+
+func checkForChangesInDebugRegex() {
 	// if  the regex string has changed ...
 	if globals.Params.DebugLogRegEx != globals.LastDebugLogRegEx {
+		lastSlashIndex := strings.LastIndex(globals.Params.DebugLogRegEx, string(os.PathSeparator))
```
Emyrk: Can you refactor this logic of separating a string into the location and regex into a function? 

I see your unit test going as far as launching a simtest, but it would be nice to have just this functionality unit tested against various strings. It would also then consistently run in circleci.

PR #774, opened 2019-06-24
Fix a case where a pokemon message was detected and still crashed the…

notworken:
… system.


Emyrk:
This bug has been showing up more recently in bond, this is a clean solution to prevent a crash :+1:

PR #778, opened 2019-06-24
Fd 821 perftest parchment

stackdump:
A load-test that  checks to make sure network progresses.

passes at 5tps 

Failes at 10
```
--- FAIL: TestPerfRebound (426.35s)
    simulation.go:163: Allocated 14 nodes
    PerfRebound_test.go:47: LLHT: 4<=>7 moved 3
    PerfRebound_test.go:50: only moved 3 blocks
```

### DIFF
--- a/longTest/PerfRebound_test.go
+++ b/longTest/PerfRebound_test.go
@@ -1,0 +1,53 @@
+package longtest
+
+import (
+	"testing"
+	"time"
+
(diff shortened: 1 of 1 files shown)


stackdump:
This test isn't needed. It seems that the net effect of this type of test was benchmarking my workstation rather than Factomd.
this was replaced w/ a test that does loading /w 5 nodes: 'LLLFF'  as part of the Dependent holding PR

PR #791, opened 2019-07-09
Fd 1091 savestate off by one

factom-clay:


Emyrk:
Verified with DBGenerator generated databases

PR #795, opened 2019-07-10
Second test at pulling changes into the Xuan branch

factablesolutions:
The bulk of these changes fall into these categories:

1) Simple adding of comments
2) changing internal 'method' names to be consistent within a file
3) removing duplicate function that was never used outside the file
4) refactoring to remove duplicate code
5) adding new tests
    - note: in log_tests.go, the new tests are for Log() and LogJSONs(), neither of which appear to be called anywhere in the code, so its arguable these functions could be deleted, since I'm not even sure if they work properly. However, I created two regressions to test them, and made the tests pass based off their current behavior (although since no one uses the tests and there were no comments, its questionable whether they are actually working properly). At any rate, they use a hard-coded line number which i don't like, since any change to the log_tests.go file will cause them to fail if the line changes. If you have any suggestions on how to fix this, i'd love to hear them.


factablesolutions:
This was also a flawed pull request, closing

PR #796, opened 2019-07-10
FD-1114 Xuan factable branch

factablesolutions:
1) Adding Comments
2) Adding tests
3) removing code duplication via refactor
4) make receivers consistent naming
etc

### DIFF
--- a/anchor/anchorRecord.go
+++ b/anchor/anchorRecord.go
@@ -17,22 +17,32 @@ import (
 	"github.com/FactomProject/factomd/common/primitives"
 )
 
-//AnchorRecord is used to construct anchor chain
+// CurrentAnchorVersion is the current anchor version
+const CurrentAnchorVersion int = 1
+
+// AnchorRecord is used to construct the anchor chain. The Factom Protocol writes an anchor into
+// a parent blockchain (Bitcoin or Ethereum) approximately every 10 minutes. The accumulated entries
+// from the previous 10 minutes are organized into a Directory Block (DBlock). The information
+// from the directory block is stored in the AnchorRecord.
 type AnchorRecord struct {
-	AnchorRecordVer int
-	DBHeight        uint32 `json:",omitempty"` // The only directory block height included in this anchor
-	KeyMR           string `json:",omitempty"` // Merkle root of the only directory block included in this anchor
+	AnchorRecordVer int    // the only supported version seems to be 1
+	DBHeight        uint32 // Factom Directory Block Height - the unique number associated with this DBlock
+	KeyMR           string // key merkle root of the directory block
 
 	DBHeightMax uint32 `json:",omitempty"` // The highest directory block height included in this anchor window
 	DBHeightMin uint32 `json:",omitempty"` // The lowest directory block height included in this anchor window
 	WindowMR    string `json:",omitempty"` // Merkle root of all directory block KeyMRs from DBHeightMin to DBHeightMax
 
-	RecordHeight uint32 // Directory block height we intended to put the AnchorRecord into
-
+	RecordHeight uint32 // This is the future DBlock height a confirmation of anchoring event X is
+	// written to (usually X+1). In principle, the confirmation status is available
+	// in the Bitcoin/Etherium structs, but RecordHeight acts as an internal syncing
+	// mechanism to ensure that all Factom servers are in sync and working on the same
+	// block (ie, not behind)
 	Bitcoin  *BitcoinStruct  `json:",omitempty"`
 	Ethereum *EthereumStruct `json:",omitempty"`
 }
 
+// BitcoinStruct contains relevant data for a Bitcoin transaction
 type BitcoinStruct struct {
 	Address     string //"1HLoD9E4SDFFPDiYfNYnkBLQ85Y51J3Zb1",
 	TXID        string //"9b0fc92260312ce44e74ef369f5c66bbb85848f2eddd5a7a1cde251e54ccfdd5", BTC Hash - in reverse byte order
@@ -41,6 +51,7 @@ type BitcoinStruct struct {
 	Offset      int32  //87
 }
 
+// EthereumStruct contains relevant data for an Ethereum transaction
 type EthereumStruct struct {
 	ContractAddress string // Address of the Ethereum anchor contract
 	TxID            string // Transaction ID of this particular anchor
@@ -52,19 +63,23 @@ type EthereumStruct struct {
 var _ interfaces.Printable = (*AnchorRecord)(nil)
 var _ interfaces.IAnchorRecord = (*AnchorRecord)(nil)
 
-func (e *AnchorRecord) JSONByte() ([]byte, error) {
-	return primitives.EncodeJSON(e)
+// JSONByte returns a []byte of the AnchorRecord encoded in Json: nil, error returned upon error
+func (ar *AnchorRecord) JSONByte() ([]byte, error) {
+	return primitives.EncodeJSON(ar)
 }
 
-func (e *AnchorRecord) JSONString() (string, error) {
-	return primitives.EncodeJSONString(e)
+// JSONString returns a string of the AnchorRecord encoded in Json: "", error returned upon error
+func (ar *AnchorRecord) JSONString() (string, error) {
+	return primitives.EncodeJSONString(ar)
 }
 
-func (e *AnchorRecord) String() string {
-	str, _ := e.JSONString()
+// String returns a string of AnchorRecord encoded in Json: "" returned upon error
+func (ar *AnchorRecord) String() string {
+	str, _ := ar.JSONString()
 	return str
 }
 
+// Marshal marshals the AnchorRecord into json format
 func (ar *AnchorRecord) Marshal() (rval []byte, err error) {
 	defer func(pe *error) {
 		if *pe != nil {
@@ -78,15 +93,17 @@ func (ar *AnchorRecord) Marshal() (rval []byte, err error) {
 	return data, nil
 }
 
+// MarshalAndSign marshals the AnchorRecord into json and signs it with the input
+// Signer, returning concatenated data of (data,signature)
 func (ar *AnchorRecord) MarshalAndSign(priv interfaces.Signer) ([]byte, error) {
-	data, err := ar.Marshal()
+	data, sigbytes, err := ar.MarshalAndSignV2(priv)
 	if err != nil {
 		return nil, err
 	}
-	sig := priv.Sign(data)
-	return append(data, []byte(fmt.Sprintf("%x", sig.Bytes()))...), nil
+	return append(data, []byte(fmt.Sprintf("%x", sigbytes))...), nil
 }
 
+// MarshalAndSignV2 marshals the AnchorRecord into json and signs, returning separate anchor and signature data
 func (ar *AnchorRecord) MarshalAndSignV2(priv interfaces.Signer) ([]byte, []byte, error) {
 	data, err := ar.Marshal()
 	if err != nil {
@@ -96,17 +113,30 @@ func (ar *AnchorRecord) MarshalAndSignV2(priv interfaces.Signer) ([]byte, []byte
 	return data, sig.Bytes(), nil
 }
 
-func (ar *AnchorRecord) Unmarshal(data []byte) error {
+// Non-exported, refactored function - is it better to use *string return for nil, or ""??
+func splitAnchorAndSignature(data []byte) (string, string, error) {
 	if len(data) == 0 {
-		return fmt.Errorf("Invalid data passed")
+		return "", "", fmt.Errorf("Invalid data passed")
 	}
 	str := string(data)
 	end := strings.LastIndex(str, "}}")
 	if end < 0 {
-		return fmt.Errorf("Found no closing bracket in `%v`", str)
+		return "", "", fmt.Errorf("Found no closing bracket in `%v`", str)
 	}
-	str = str[:end+2]
-	err := json.Unmarshal([]byte(str), ar)
+	// Signature comes after anchor
+	anchorStr := str[:end+2]
+	sigStr := str[end+2:]
+
+	return anchorStr, sigStr, nil
+}
+
+// Unmarshal unmarshals json format input into this AnchorRecord
+func (ar *AnchorRecord) Unmarshal(data []byte) error {
+	str, _, err := splitAnchorAndSignature(data)
+	if err != nil {
+		return err
+	}
+	err = json.Unmarshal([]byte(str), ar)
 	if err != nil {
 		return err
 	}
@@ -114,6 +144,18 @@ func (ar *AnchorRecord) Unmarshal(data []byte) error {
 	return nil
 }
 
+// IsSame returns true iff all the fields of ar==ar2
+func (ar *AnchorRecord) IsSame(ar2 *AnchorRecord) bool {
+	if ar.AnchorRecordVer != ar2.AnchorRecordVer || ar.DBHeight != ar2.DBHeight || ar.KeyMR != ar2.KeyMR ||
+		ar.DBHeightMax != ar2.DBHeightMax || ar.DBHeightMin != ar2.DBHeightMin || ar.WindowMR != ar2.WindowMR ||
+		ar.RecordHeight != ar2.RecordHeight || (ar.Bitcoin != nil && ar.Bitcoin.IsSame(ar2.Bitcoin) == false) ||
+		(ar.Ethereum != nil && ar.Ethereum.IsSame(ar2.Ethereum) == false) {
+		return false
+	}
+	return true
+}
+
+// UnmarshalAnchorRecord unmarshals a json format input into a new AnchorRecord
 func UnmarshalAnchorRecord(data []byte) (*AnchorRecord, error) {
 	ar := new(AnchorRecord)
 	err := ar.Unmarshal(data)
@@ -123,47 +165,50 @@ func UnmarshalAnchorRecord(data []byte) (*AnchorRecord, error) {
 	return ar, nil
 }
 
-func UnmarshalAndValidateAnchorRecord(data []byte, publicKeys []interfaces.Verifier) (*AnchorRecord, bool, error) {
-	if len(data) == 0 {
-		return nil, false, fmt.Errorf("Invalid data passed")
+// verifyAnchorAndSignature verifies the data and signature from the public keys - unexported
+func verifyAnchorAndSignature(data []byte, sig *primitives.ByteSliceSig, publicKeys []interfaces.Verifier) (bool, error) {
+	fixed, err := sig.GetFixed()
+	if err != nil {
+		return false, err
 	}
-	str := string(data)
-	end := strings.LastIndex(str, "}}")
-	if end < 0 {
-		return nil, false, fmt.Errorf("Found no closing bracket in `%v`", str)
-	}
-	anchorStr := str[:end+2]
-	signatureStr := str[end+2:]
 
-	sig := new(primitives.ByteSliceSig)
-	err := sig.UnmarshalText([]byte(signatureStr))
+	valid := false
+	for _, publicKey := range publicKeys {
+		valid = publicKey.Verify(data, &fixed)
+		if valid == true {
+			break
+		}
+	}
+	return valid, nil
+}
+
+// UnmarshalAndValidateAnchorRecord unmarshals signed json data and verifies signature with public keys
+func UnmarshalAndValidateAnchorRecord(data []byte, publicKeys []interfaces.Verifier) (*AnchorRecord, bool, error) {
+	anchorStr, signatureStr, err := splitAnchorAndSignature(data)
 	if err != nil {
 		return nil, false, err
 	}
-	fixed, err := sig.GetFixed()
+
+	sig := new(primitives.ByteSliceSig)
+	err = sig.UnmarshalText([]byte(signatureStr))
 	if err != nil {
 		return nil, false, err
 	}
 
 	valid := false
-	for _, publicKey := range publicKeys {
-		valid = publicKey.Verify([]byte(anchorStr), &fixed)
-		if valid == true {
-			break
-		}
-	}
+	valid, err = verifyAnchorAndSignature([]byte(anchorStr), sig, publicKeys)
 	if valid == false {
-		return nil, false, nil
+		return nil, false, err
 	}
 
-	ar := new(AnchorRecord)
-	err = ar.Unmarshal(data)
+	ar, err := UnmarshalAnchorRecord(data)
 	if err != nil {
 		return nil, false, err
 	}
 	return ar, true, nil
 }
 
+// UnmarshalAndValidateAnchorRecordV2 unmarshals json data and verifies external signature with public keys
 func UnmarshalAndValidateAnchorRecordV2(data []byte, extIDs [][]byte, publicKeys []interfaces.Verifier) (*AnchorRecord, bool, error) {
 	if len(data) == 0 {
 		return nil, false, fmt.Errorf("Invalid data passed")
@@ -174,20 +219,9 @@ func UnmarshalAndValidateAnchorRecordV2(data []byte, extIDs [][]byte, publicKeys
 
 	sig := new(primitives.ByteSliceSig)
 	sig.UnmarshalBinary(extIDs[0])
-	fixed, err := sig.GetFixed()
-	if err != nil {
-		return nil, false, err
-	}
-
-	valid := false
-	for _, publicKey := range publicKeys {
-		valid = publicKey.Verify(data, &fixed)
-		if valid == true {
-			break
-		}
-	}
+	valid, err := verifyAnchorAndSignature(data, sig, publicKeys)
 	if valid == false {
-		return nil, false, nil
+		return nil, false, err
 	}
 
 	ar := new(AnchorRecord)
@@ -198,6 +232,7 @@ func UnmarshalAndValidateAnchorRecordV2(data []byte, extIDs [][]byte, publicKeys
 	return ar, true, nil
 }
 
+// UnmarshalAndValidateAnchorEntryAnyVersion unmarshals json data and verifies with either internal or external signatures against public keys
 func UnmarshalAndValidateAnchorEntryAnyVersion(entry interfaces.IEBEntry, publicKeys []interfaces.Verifier) (*AnchorRecord, bool, error) {
 	ar, valid, err := UnmarshalAndValidateAnchorRecord(entry.GetContent(), publicKeys)
 	if ar == nil {
@@ -207,11 +242,30 @@ func UnmarshalAndValidateAnchorEntryAnyVersion(entry interfaces.IEBEntry, public
 	return ar, valid, err
 }
 
+// CreateAnchorRecordFromDBlock creates new AnchorRecord from dBlock
 func CreateAnchorRecordFromDBlock(dBlock interfaces.IDirectoryBlock) *AnchorRecord {
 	ar := new(AnchorRecord)
-	ar.AnchorRecordVer = 1
+	ar.AnchorRecordVer = CurrentAnchorVersion
 	ar.DBHeight = dBlock.GetHeader().GetDBHeight()
 	ar.KeyMR = dBlock.DatabasePrimaryIndex().String()
 	ar.RecordHeight = ar.DBHeight
 	return ar
+}
+
+// IsSame returns true iff all fields of BitcoinStructs bc==bc2
+func (bc *BitcoinStruct) IsSame(bc2 *BitcoinStruct) bool {
+	if bc.Address != bc2.Address || bc.TXID != bc2.TXID || bc.BlockHeight != bc2.BlockHeight || bc.BlockHash != bc2.BlockHash ||
+		bc.Offset != bc2.Offset {
+		return false
+	}
+	return true
+}
+
+// IsSame returns true iff all fields of EthereumStructs es==es2
+func (es *EthereumStruct) IsSame(es2 *EthereumStruct) bool {
+	if es.ContractAddress != es2.ContractAddress || es.TxID != es2.TxID || es.BlockHeight != es2.BlockHeight || es.BlockHash != es2.BlockHash ||
+		es.TxIndex != es.TxIndex {
+		return false
+	}
+	return false
 }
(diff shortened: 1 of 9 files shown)


anchor/anchorRecord.go
```diff
@@ -96,24 +113,49 @@ func (ar *AnchorRecord) MarshalAndSignV2(priv interfaces.Signer) ([]byte, []byte
...
+	}
+	err = json.Unmarshal([]byte(str), ar)
 	if err != nil {
 		return err
 	}
 
 	return nil
 }
 
+// IsSame returns true iff all the fields of ar==ar2
```
Emyrk: `(ar.Ethereum != nil && ar.Ethereum.IsSame(ar2.Ethereum) == false)`
So if `ar.Etheruem` is nil, and `ar2.Etheruem` is not nil, then this would report it as being the same?
If 1 has a nil, I'd suggest checking the other is also nil. This goes for the Bitcoin field too

common/primitives/log.go
```diff
@@ -12,6 +12,8 @@ import (
 	"strconv"
 )
 
+// Log prints the formated string with args (using fmt.Printf) with a prepended file and line number of
+// the calling function: "<file>:<linenumber> - <fmt.Printf_output>\n"
 func Log(format string, args ...interface{}) {
```
Emyrk: All this log stuff isn't used. We changed how we log, so honestly we could delete this file

anchor/anchorRecord.go
factablesolutions: Good catch, I had intended that IsSame() would return proper false if nil is passed in, however, in looking at my implementation, it will crash. I think the correct fix is to change the bitcoin/ethereum IsSame() function to properly return false when nil is passed in. Right now it will crash. :-\

common/primitives/log.go
factablesolutions: I'd be in favor of deleting the functions if they are no longer used anywhere or relevant. But I let you guys make that determination.

anchor/anchorRecord.go
```diff
@@ -17,22 +17,32 @@ import (
...
 	DBHeightMin uint32 `json:",omitempty"` // The lowest directory block height included in this anchor window
 	WindowMR    string `json:",omitempty"` // Merkle root of all directory block KeyMRs from DBHeightMin to DBHeightMax
 
-	RecordHeight uint32 // Directory block height we intended to put the AnchorRecord into
-
+	RecordHeight uint32 // This is the future DBlock height a confirmation of anchoring event X is
+	// written to (usually X+1). In principle, the confirmation status is available
+	// in the Bitcoin/Etherium structs, but RecordHeight acts as an internal syncing
+	// mechanism to ensure that all Factom servers are in sync and working on the same
+	// block (ie, not behind)
```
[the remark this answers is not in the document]
factablesolutions: Yeah, this was confusing for me. I had a conversation with Sam on 6/24 in developer discussion on discord and this was my interpretation of what I understood. I'm happy to make any changes requested, or just delete my comments on this particular member completely if its wrong. But I think it needs a comment that is correct, because the parameters usefulness isn't readily apparent.

anchor/anchorRecord.go
```diff
@@ -96,24 +113,49 @@ func (ar *AnchorRecord) MarshalAndSignV2(priv interfaces.Signer) ([]byte, []byte
 	return data, sig.Bytes(), nil
 }
 
-func (ar *AnchorRecord) Unmarshal(data []byte) error {
+// Non-exported, refactored function - is it better to use *string return for nil, or ""??
```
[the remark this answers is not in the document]
factablesolutions: I kept the function as it is, and changed the comment

anchor/anchorRecord.go
factablesolutions: I updated this as well to fix the nil input, and found a bad return

anchor/anchorRecord.go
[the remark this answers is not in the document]
sambarnes: The usefulness of the parameter isn't readily apparent because it isn't actually all that useful :) I only kept it in here for backwards compatibility reasons since it was already here for bitcoin anchors. We will likely remove it in the next iteration of anchoring when we revamp both btc and eth systems to be more decentralized. Steven is right, anchoring has nothing to do with consensus at this point in time.

As for the comment, I'd adapt it to be more like the following:
```
// Likely to be deprecated soon. This is the future DBlock height a confirmation of anchoring event X is
// **intended** to be written to (usually X+1). The field is fairly useless since you can't know for sure
// what block an entry will be included in until after the entry is confirmed, and by that point, you can
// get the block height from it's parent Entry Block.
```

anchor/anchorRecord.go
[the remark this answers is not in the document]
sambarnes: I didn't design v1 or v2 Anchor Records, simply added more fields to v2 to support Ethereum as well. 

But the explanation is that in V1 (for some reason), they chose to not utilize external IDs for the signature. So the content field for a ton of our bitcoin anchor records looks like this:
```
{"AnchorRecordVer":1,"DBHeight":70405,"KeyMR":"2c6c6247ae2602a08d5b59850682a60228bbc23ae66c67916836b684fc8daea9","RecordHeight":70405,"Bitcoin":{"Address":"1K2SXgApmo9uZoyahvsbSanpVWbzZWVVMF","TXID":"83ea4629b0e5d7ae7ee313d377dc4f9cb258652c76937afce4d0be29add6d8a3","BlockHeight":446482,"BlockHash":"00000000000000000166d612d5595e2b1cd88d71d695fc580af64d8da8658c23","Offset":1309}}98bb464b7b0cfd2cd43091e519193d62fc60909298761e512d34c2517f80f70eb07d5aa06ced19a85b0507cf09836141b328eb88e046bda5de021b14281a910e
```
https://explorer.factom.com/chains/df3ade9eec4b08d5379cc64270c30ea7315d8a8a1a69efe2b98a60ecdd69e604/entries/e47fe17ea16474444d3895d6048b2ade4c71114f9742d31a6e1d7d035019e2ee

They moved away from this in v2 and started using the ext-ids and content fields more sanely. But alas, immutability forces us to support old code that might not be as pretty as we'd have hoped.

Maybe a comment detailing how this is for Anchor Record V1 support, and should not be used otherwise

anchor/anchorRecord.go
```diff
@@ -17,22 +17,32 @@ import (
 	"github.com/FactomProject/factomd/common/primitives"
 )
 
-//AnchorRecord is used to construct anchor chain
+// CurrentAnchorVersion is the current anchor version
+const CurrentAnchorVersion int = 1
```
sambarnes: We're actually using anchor record V2. You can tell because v1 has signature in the content field, and v2 has it in the external ids. The only caveat is that whoever wrote the bitcoin anchor writer did not update the version in their json. As you can see in this recent bitcoin entry:
https://explorer.factom.com/chains/df3ade9eec4b08d5379cc64270c30ea7315d8a8a1a69efe2b98a60ecdd69e604/entries/002377b0381f4eb48df0deb4025a6d98323434d60ff999fb29a27939e9177a04

Ethereum anchors use to correct version 2 in the json though:
https://explorer.factom.com/chains/6e4540d08d5ac6a1a394e982fb6a2ab8b516ee751c37420055141b94fe070bfe/entries/45af639e22d2c557f6f01fa6492acd45667ae63f4d9dcc97e18b7bba8e600e0f

Good catch though. I'll make a ticket to change that part of the bitcoin anchoring code.

I don't know how much value the constant provides here though, since we'll always need to be able to unmarshal v1 and v2 when nodes are syncing

anchor/anchorRecord.go
```diff
@@ -17,22 +17,32 @@ import (
...
+
+// AnchorRecord is used to construct the anchor chain. The Factom Protocol writes an anchor into
+// a parent blockchain (Bitcoin or Ethereum) approximately every 10 minutes. The accumulated entries
+// from the previous 10 minutes are organized into a Directory Block (DBlock). The information
+// from the directory block is stored in the AnchorRecord.
 type AnchorRecord struct {
-	AnchorRecordVer int
-	DBHeight        uint32 `json:",omitempty"` // The only directory block height included in this anchor
-	KeyMR           string `json:",omitempty"` // Merkle root of the only directory block included in this anchor
+	AnchorRecordVer int    // the only supported version seems to be 1
```
sambarnes: as noted in another comment, version 2 exists as well

anchor/anchorRecord.go
```diff
@@ -207,11 +242,30 @@ func UnmarshalAndValidateAnchorEntryAnyVersion(entry interfaces.IEBEntry, public
 	return ar, valid, err
 }
 
+// CreateAnchorRecordFromDBlock creates new AnchorRecord from dBlock
 func CreateAnchorRecordFromDBlock(dBlock interfaces.IDirectoryBlock) *AnchorRecord {
```
sambarnes: This function isn't used anywhere, I'd just delete it and all its references in the tests

sambarnes:
My bad for not bundling the above comments into a single review. But as a last comment, have you booted this against mainnet and verified that all bitcoin and ethereum anchors are being parsed correctly? You can do it with an existing node:

```factomd --reparseanchorchains```

then send the request for some varying heights:
```
curl -X POST --data-binary '{"jsonrpc": "2.0", "id": 0, "method": "anchors", "params":{"height":190000}}' -H 'content-type:text/plain;' http://localhost:8088/v2
```

File: anchor/anchorRecord.go (hunk 1 above)
Comment (Emyrk) [reply anchor/anchorRecord.go]: I see how you resolved it.

**Comment by factablesolutions** [comment]
I'm running factomd --reparseanchorchains on my local computer. Its taking a long time because i think it has to download the block chain?:
....
              FNode0 Saving C:\Users\mlam/.factom/m2/FastBoot_MAIN_v11.db for dbht 57001
              FNode0 Saving C:\Users\mlam/.factom/m2/FastBoot_MAIN_v11.db for dbht 58001
              FNode0 Saving C:\Users\mlam/.factom/m2/FastBoot_MAIN_v11.db for dbht 59001

**Comment by Emyrk** [comment]
> I'm running factomd --reparseanchorchains on my local computer. Its taking a long time because i think it has to download the block chain?:
> ....
> FNode0 Saving C:\Users\mlam/.factom/m2/FastBoot_MAIN_v11.db for dbht 57001
> FNode0 Saving C:\Users\mlam/.factom/m2/FastBoot_MAIN_v11.db for dbht 58001
> FNode0 Saving C:\Users\mlam/.factom/m2/FastBoot_MAIN_v11.db for dbht 59001

Yes you will need to sync up mainnet. I don't think there is any real feedback from reparseanchorchains. When sync factomd, you will also sync anchors. Sam told you to resync them, as that is how we can test the anchor syncing without needing to do a full sync.

You can check the anchors at the heights you synced though with the curl that same posted, and just change the height param to a height you have.

File: anchor/anchorRecord.go (hunk 3 above)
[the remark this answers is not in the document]
Comment (factablesolutions) [reply anchor/anchorRecord.go]: I replaced my comments with your comment suggestions above

File: anchor/anchorRecord.go (hunk 5 above)
Comment (factablesolutions) [reply anchor/anchorRecord.go]: Ah, I thought V2 indicated a second version of the function, rather than a different version of AnchorRecord. I added clarifying comments about v1 vs v2. I also modified the anchor tests to use v2 where applicable. Except for TestValidateAnchorRecordV2, this test fails if I move the anchor string to use version 2 instead of version 1, so I left it. I'm not yet sure why it fails. I also removed the const Version I put in

File: anchor/anchorRecord.go (hunk 6 above)
Comment (factablesolutions) [reply anchor/anchorRecord.go]: Fixed this comment

File: anchor/anchorRecord.go (hunk 7 above)
Comment (factablesolutions) [reply anchor/anchorRecord.go]: I deleted CreateAnchorRecordFromDBlock() and removed it from the files it was used, just duplicating a small amount of code in their respective locations

**Comment by factablesolutions** [comment]
I'm still working on the --reparseanchorchains, it seems to stall out at 60956. The control panel says its 100% sync'd, but that seems weirdly low. I'm not sure if this indicates something is wrong with my build, or whether my system is just not set up to run the command properly.

**Comment by Emyrk** [comment]
> I'm still working on the --reparseanchorchains, it seems to stall out at 60956. The control panel says its 100% sync'd, but that seems weirdly low. I'm not sure if this indicates something is wrong with my build, or whether my system is just not set up to run the command properly.

Did your computer go to sleep or anything that could interrupt it?

**Comment by Emyrk** [comment]
Also @factablesolutions can you sign the CLA file in the root directory?

**Comment by factablesolutions** [comment]
See comment 2 above, I signed the CLA yesterday. Unless I did something wrong?

**Comment by Emyrk** [comment]
> See comment 2 above, I signed the CLA yesterday. Unless I did something wrong?

No you did it right, I just had this PR open from yesterday and didn't refresh the page. Sorry

**Comment by factablesolutions** [comment]
Ok, so i believe I've satisfied all your requests except for the --reparseanchorchains. For some reason when i run it, it only gets to:
....
Reparsing anchor chains...
              FNode0 Loading blocks from disk. Database load going from 59000 (savestate) to 60956 (disk)
Using NetworkName "MAIN"
Starting API server
Starting Control Panel on http://localhost:8090/
              FNode0 Loading Block   60000 / 60956. Blocks per second    37.46 average bps    37.43 Progress 0:26 remaining 0:25 Estimated Total Time: 0:51
              FNode0 Loading complete 60956.
              FNode0 Saving C:\Users\mlam/.factom/m2/FastBoot_MAIN_v11.db for dbht 60000

Is there a way to clear out the files? can i just delete the .factom directory? or the m2 directory? And i can rerun. Also the --reparseanchorchains doesn't seem to tell me when its done, so i imagine that just continually runs factomd, and I should use a separate window to run the curl command on a separate terminal?

**Comment by factablesolutions** [comment]
Yeah, I'm going to need some help on how to fix this. I deleted my m2 directory, and reran, but not i get:
Starting Control Panel on http://localhost:8090/
              FNode0 Loading complete 0.
And the control panel says 100% sync'd with 0 blocks. So something is screwed up with running it on my local computer. Not sure how to get it to download the block chain. I tried just running ./factomd and it does the same thing, stops at 0 block height

**Comment by Emyrk** [comment]
> Ok, so i believe I've satisfied all your requests except for the --reparseanchorchains. For some reason when i run it, it only gets to:
> ....
> Reparsing anchor chains...
> FNode0 Loading blocks from disk. Database load going from 59000 (savestate) to 60956 (disk)
> Using NetworkName "MAIN"
> Starting API server
> Starting Control Panel on http://localhost:8090/
> FNode0 Loading Block 60000 / 60956. Blocks per second 37.46 average bps 37.43 Progress 0:26 remaining 0:25 Estimated Total Time: 0:51
> FNode0 Loading complete 60956.
> FNode0 Saving C:\Users\mlam/.factom/m2/FastBoot_MAIN_v11.db for dbht 60000
> 
> Is there a way to clear out the files? can i just delete the .factom directory? or the m2 directory? And i can rerun. Also the --reparseanchorchains doesn't seem to tell me when its done, so i imagine that just continually runs factomd, and I should use a separate window to run the curl command on a separate terminal?

Reparse chains doesn't actually give you any feedback. If the control panel is launched, than the parsing completed without panicing. So you should be able to run the api call to retrieve the anchors for up to your blkheight (60k)

As for the 100% synced, this is expected. There has been some height reworks which unfortunately affected the api sync status. It has to do with some security related patches, and a solution for the control panel has not been prioritized.

**Comment by factablesolutions** [comment]
So I've restarted the --reparseanchorchains run, its pretty slow showing 73k block height in about 6 hours. Also, I'm on a windows machine for my development for now. Is there an analogue for the curl command on windows I can run? I'm not familiar with curl but it looks like it I'm just trying get the code to execute a function via the json rpc.

**Comment by factablesolutions** [comment]
Ok, I downloaded curl for windows. I'll have to play around with it some, the command as Sam supplied is throwing some errors, but it may be a windows/linux issue. I'll have to go through the documentation tomorrow to figure out how to use curl:

curl.exe -X POST --data-binary '{"jsonrpc": "2.0", "id": 0, "method": "anchors", "params":{"height":19}}' -H 'content-type:text/plain;' http://localhost:8088/v2
curl: (6) Could not resolve host: 2.0,
curl: (6) Could not resolve host: id
curl: (6) Could not resolve host: 0,
curl: (6) Could not resolve host: method
curl: (6) Could not resolve host: anchors,
curl: (3) unmatched close brace/bracket in URL position 18:
params:{height:19}}'
                 ^
{"jsonrpc":"2.0","id":null,"error":{"code":-32600,"message":"Invalid Request"}}

File: anchor/anchorRecord.go [hunk 8]
```diff
@@ -17,22 +17,29 @@ import (
...
+// a parent blockchain (Bitcoin or Ethereum) approximately every 10 minutes. The accumulated entries
+// from the previous 10 minutes are organized into a Directory Block (DBlock). The information
+// from the directory block is stored in the AnchorRecord.
 type AnchorRecord struct {
-	AnchorRecordVer int
-	DBHeight        uint32 `json:",omitempty"` // The only directory block height included in this anchor
-	KeyMR           string `json:",omitempty"` // Merkle root of the only directory block included in this anchor
+	AnchorRecordVer int    // version 1 places signature content in the field, version 2 uses external IDs for signature
+	DBHeight        uint32 // Factom Directory Block Height - the unique number associated with this DBlock
+	KeyMR           string // key merkle root of the directory block
```
PaulBernier: Does the removal of `json:",omitempty"` change the behavior here? If yes, why is safe and what's the benefit of the change? If not, why apply this removal to only those 2 fields and not the rest?

WhoSoup:
> I'll have to go through the documentation tomorrow to figure out how to use curl:

windows doesn't recognize ' as delimiters, you'll have to use " and escape quotes inside, ie: 
`curl.exe -X POST --data-binary "{\"jsonrpc\": \"2.0\", \"id\": 0, \"method\": \"heights\", \"params\":{}}" -H "content-type:text/plain;" http://localhost:8088/v2`

anchor/anchorRecord.go
```diff
@@ -96,24 +110,49 @@ func (ar *AnchorRecord) MarshalAndSignV2(priv interfaces.Signer) ([]byte, []byte
 	return data, sig.Bytes(), nil
 }
 
-func (ar *AnchorRecord) Unmarshal(data []byte) error {
+// Non-exported, refactored function that splits the AnchorRecord and its signature
```
PaulBernier: Is specifying "Non-exported" in the comment necessary? I don't think it conveys any additional information to the reader.

common/primitives/key.go
```diff
@@ -143,54 +148,62 @@ func (c *PublicKey) Copy() (*PublicKey, error) {
...
 		if *pe != nil {
 			fmt.Fprintf(os.Stderr, "PublicKey.MarshalText err:%v", *pe)
 		}
 	}(&err)
-	return []byte(pk.String()), nil
+	return []byte(k.String()), nil
 }
 
-func (pk *PublicKey) UnmarshalText(b []byte) error {
+// UnmarshalText decodes the input []byte public key into 'pk'
```
PaulBernier: Not true anymore, because you renamed pk => k :smile:

anchor/anchorRecord.go
```diff
@@ -123,47 +162,50 @@ func UnmarshalAnchorRecord(data []byte) (*AnchorRecord, error) {
 	return ar, nil
 }
 
-func UnmarshalAndValidateAnchorRecord(data []byte, publicKeys []interfaces.Verifier) (*AnchorRecord, bool, error) {
-	if len(data) == 0 {
-		return nil, false, fmt.Errorf("Invalid data passed")
+// verifyAnchorAndSignature verifies the data and signature from the public keys - unexported
```
PaulBernier: same here, "unexported" seems unecessary,

anchor/anchorRecord.go
```diff
@@ -207,11 +239,20 @@ func UnmarshalAndValidateAnchorEntryAnyVersion(entry interfaces.IEBEntry, public
...
 }
 
-func CreateAnchorRecordFromDBlock(dBlock interfaces.IDirectoryBlock) *AnchorRecord {
-	ar := new(AnchorRecord)
-	ar.AnchorRecordVer = 1
-	ar.DBHeight = dBlock.GetHeader().GetDBHeight()
-	ar.KeyMR = dBlock.DatabasePrimaryIndex().String()
-	ar.RecordHeight = ar.DBHeight
-	return ar
+// IsSame returns true iff all fields of BitcoinStructs bc==bc2
```
PaulBernier: typo `iff`

common/primitives/json.go
```diff
@@ -56,43 +68,58 @@ func ParseJSON2Request(request string) (*JSON2Request, error) {
...
 	j := new(JSON2Response)
 	j.JSONRPC = "2.0"
 	return j
 }
 
+// AddError sets the member Error with a new JSONError formed from the inputs
 func (j *JSON2Response) AddError(code int, message string, data interface{}) {
-	e := NewJSONError(code, message, data)
-	j.Error = e
+	err := NewJSONError(code, message, data)
```
PaulBernier: why this intermediary variable `err` at all then :)

anchor/anchorRecord.go
factablesolutions: Ahh! Good catch, this was an error on merging. Sam had made some changes and I screwed that up. Fixed

anchor/anchorRecord.go
factablesolutions: Clarified comments to what I had intended, that its non-exported for now because it was created during refactor, but there's no reason it couldn't be exported in the future.

common/primitives/key.go
factablesolutions: Fixed!

anchor/anchorRecord.go
factablesolutions: Same clarifying comment as above, fixed to say it could be exported in the future.

anchor/anchorRecord.go
factablesolutions: iff = if and only if , it was already in use in the file, so I continued its usage, but happy to clean up to be more explicit if necessary

common/primitives/json.go
factablesolutions: good catch, i missed that in clean up. :-)

factablesolutions:
> windows console doesn't recognize ' as delimiters, you'll have to use " and escape quotes inside, ie:
> `curl.exe -X POST --data-binary "{\"jsonrpc\": \"2.0\", \"id\": 0, \"method\": \"heights\", \"params\":{}}" -H "content-type:text/plain;" http://localhost:8088/v2`

Thanks Who. Curl runs now, and spits out a whole lot of data:

<!doctype html>
<html class="no-js" lang="en" dir="ltr">
    <head>
        <meta charset="utf-8">
        <meta http-equiv="x-ua-compatible" content="ie=edge">
        <meta name="viewport" content="width=device-width, initial-scale=1.0">
        <link rel="shortcut icon" type="image/x-icon" href="img/icon/favicon.ico">
        <title>Control Panel - Factom Foundation</title>
        <link rel="stylesheet" href="css/foundation.min.css">
        <link rel="stylesheet" href="css/font-awesome.min.css">
        <link rel="stylesheet" href="css/app.css">
.... whole lot more stuff.....

I downloaded the jq-win64.exe file, and when I try to use it with the '|' in Sam's command, I'm also getting errors:
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0parse error: Invalid numeric literal at line 3, column 10
100  9533    0  9475  100    58  87731    537 --:--:-- --:--:-- --:--:-- 88268
curl: (23) Failed writing body (1480 != 5805)

I've also never used jq, so if you have thoughts, let me know.

anchor/anchorRecord.go
factablesolutions: Note, i also fixed the DBHeight as well, which i had also merged poorly.

WhoSoup:
> Thanks Who. Curl runs now, and spits out a whole lot of data:

that looks like the wrong port. localhost:8088 is the default factomd api (the one you want), :8090 is the control panel

anchor/anchorRecord.go
WhoSoup: agreed with paul, golang is self-documenting in that way and saying that it's not exported in a comment is not needed. documenting that a function was changed from exported to unexported should be in the commit log or change log or somewhere else (if it's an important change that affects other packages)

i would personally comment it as something like `// helper function that splits the AnchorRecord and its signature`

factablesolutions:
Good catch Who. I have it running,but its returning invalid request, even when i switch back to the FD-851 branch without my changes. So here's exactly what I'm doing. In one cmd window, I'm running factomd --reparseanchorchains. I wait for it to get established, and then in a separate cmd window I'm running the curl command below:

>curl.exe -X POST --data-binary "{"jsonrpc": "2.0", "id": 0, "method": "anchors", "params":{"height":1}}" -H "content-type:text/plain;" http://localhost:8088/v2 | jq-win64.exe
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   136  100    79  100    57   1612   1163 --:--:-- --:--:-- --:--:--  2775
{
  "jsonrpc": "2.0",
  "id": null,
  "error": {
    "code": -32600,
    "message": "Invalid Request"
  }
}

Code seems to run now, but returns invalid request.

anchor/anchorRecord.go
factablesolutions: My comment was really supposed to be why its non-exported. And the reason is just that I chose not to export it. Perhaps I'm thinking in too much of C++ way, but private methods in C++ are kept private as a way to control the exposed interface. So you often think carefully about what methods you expose to the outside world. My intended purpose for the comment is just that there is no good reason to have it unexported other than it was refactored an therefore is only seen internally right now and so its unexported. There's no reason it couldn't be exported in the future in the future if the functionality is useful to outside parties (not sure it is or not). At any rate. I can happily change the comment to drop all mention of non-exported, since I don't think its a big deal either way.

Emyrk: I'm in favor of dropping the comment to something like @WhoSoup suggested. Documenting the obvious makes a change later now also need to touch the comment. Since the function is upper case, the comment doesn't actually stop someone from using it externally.

So if it is possible to make private, lowercasing the function method is fine. If you need it exported (for a unit test or w/e), it's fine as well.

anchor/anchorRecord.go
Emyrk: iff is fine :+1:

Emyrk:
> Good catch Who. I have it running,but its returning invalid request, even when i switch back to the FD-851 branch without my changes. So here's exactly what I'm doing. In one cmd window, I'm running factomd --reparseanchorchains. I wait for it to get established, and then in a separate cmd window I'm running the curl command below:
> 
> > curl.exe -X POST --data-binary "{"jsonrpc": "2.0", "id": 0, "method": "anchors", "params":{"height":1}}" -H "content-type:text/plain;" http://localhost:8088/v2 | jq-win64.exe
> > % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
> > Dload  Upload   Total   Spent    Left  Speed
> > 100   136  100    79  100    57   1612   1163 --:--:-- --:--:-- --:--:--  2775
> > {
> > "jsonrpc": "2.0",
> > "id": null,
> > "error": {
> > "code": -32600,
> > "message": "Invalid Request"
> > }
> > }
> 
> Code seems to run now, but returns invalid request.

You cannot use the double quotes to wrap your post data and use the double quotes inside.

```
curl.exe -X POST --data-binary '{"jsonrpc": "2.0", "id": 0, "method": "anchors", "params":{"height":1}}' -H "content-type:text/plain;" http://localhost:8088/v2 | jq-win64.exe
```

anchor/anchorRecord.go
factablesolutions: I removed the export comment

anchor/anchorRecord.go
factablesolutions: I removed the export comment

anchor/anchorRecord.go
```diff
@@ -52,19 +60,23 @@ type EthereumStruct struct {
 var _ interfaces.Printable = (*AnchorRecord)(nil)
 var _ interfaces.IAnchorRecord = (*AnchorRecord)(nil)
 
-func (e *AnchorRecord) JSONByte() ([]byte, error) {
-	return primitives.EncodeJSON(e)
+// JSONByte returns a []byte of the AnchorRecord encoded in Json: nil, error returned upon error
```
Emyrk: We should remove the `: nil, error returned upon error`

It's obvious in the return fields.

common/primitives/key.go
```diff
@@ -210,6 +225,7 @@ func (k *PublicKey) UnmarshalBinaryData(p []byte) ([]byte, error) {
 	return p[ed25519.PublicKeySize:], nil
 }
 
+// UnmarshalBinary unmarshals the input p into the public key k, returning an error if needed
```
Emyrk: Remove `returning an error if needed`

common/primitives/key.go
```diff
@@ -202,6 +215,8 @@ func (k *PublicKey) MarshalBinary() (rval []byte, err error) {
 	return buf.DeepCopyBytes(), nil
 }
 
+// UnmarshalBinaryData unmarshals the first section of input p of the size of a public key, and returns
+// the residual data p along with an error if needed
```
Emyrk: Remove along with an error if needed

common/primitives/log.go
```diff
@@ -48,6 +60,7 @@ func EncodeJSON(data interface{}) ([]byte, error) {
 	return encoded, nil
 }
 
+// EncodeJSONString marshals data into Json format and returns a string
```
Emyrk: Remove `returns a string`

anchor/anchorRecord.go
factablesolutions: done

common/primitives/key.go
factablesolutions: done

common/primitives/key.go
factablesolutions: done

common/primitives/log.go
factablesolutions: done

factablesolutions:
FYI, the issue was I needed to escape each " inside the outer " in windows. This allows curl to run properly

PR #799, opened 2019-07-12
Fd 932 run all tests in scheduled job

stackdump:
NOTE: still 3 tests failing - but want to get opinion on this type of testing setup.

The Implication of this is that we'd isolate tests by built tags, so if there are groups of specific tests to run then we'd get them into source files w/ proper tagging & you can then use `go test --list` to build the runlist


stackdump:
To make this work inside golang had to configure build tags
https://gist.github.com/stackdump/8caeda7cf9ec4e5ccc7557f62a805cc3#file-factomd-iml-L4-L12

stackdump:
NOTE: config is done here 
image

Get to this window by using preferences 
image

Emyrk:
I'm having issues running `go test` from the cmd with these tags. It's preventing me from running the tests

stackdump:
We spoke about this during standup a bit - 
Because adding tags then requires them to be specified - we will not adopt build tags as a means of separating tests - instead will finish relocating sim tests into ./simTest & ./longTest folders

additional sub groupings will be possible by following the same naming convention for names of tests

see existing examples from @factom-clay 
```
engine/factomd_long_test.go:func TestDBsigElectionEvery2Block_long(t *testing.T) {
engine/factomd_long_test.go:func TestGrants_long(t *testing.T) {
engine/factomd_long_test.go:func TestTestNetCoinBaseActivation_long(t *testing.T) {
engine/runSimTest.sh:    pattern='(?<!_long)$'
```

NOTE: will relocate these tests out of engine

stackdump:
Replaced this PR with https://github.com/FactomProject/factomd/pull/802

PR #801, opened 2019-07-15
Fd 450 remove hoisie

lvandenbrink:


wsapi/debugapi.go
```diff
@@ -7,6 +7,7 @@ package wsapi
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/prometheus/common/log"
```
Emyrk: Is there a reason why we are adding a new logging package? Will logrus work? 
We already have logrus in other parts, and this log package is not building as it's dependencies are not in the glide file.

Emyrk:
Update the glide.lock to include gorilla mux

PR #804, opened 2019-07-16
Fd 1114 factable code cleanup

Emyrk:


Emyrk:
Continued from https://github.com/FactomProject/factomd/pull/796#commits-pushed-a05e9c0

PR #806, opened 2019-07-17
FD 1124 BalanceHash Error rebase to Bond

Emyrk:
Correct double process of blocks on boot


Emyrk:
Not applying this to bond

PR #813, opened 2019-07-23
Fd 1089 extend setupsim

stackdump:
Tweak SimSetup to allow Fnode0 to be demoted and replaced as a Follower
This also swaps the identity - net gain is to allow a test where fnode0 isn't special

### DIFF
--- a/testHelper/simulation.go
+++ b/testHelper/simulation.go
@@ -159,6 +159,22 @@ func setTestTimeouts(state0 *state.State, calcTime time.Duration) {
 	fmt.Printf("Starting timeout timer:  Expected test to take %s or %d blocks\n", calcTime.String(), ExpectedHeight)
 }
 
+func isDefaultSim(givenNodes string) bool {
+	nodeList := []rune(givenNodes)
+
+	if nodeList[0] != 'L' {
+		return false
+	}
+
+	for x := 1; x < len(givenNodes); x++ {
+		if nodeList[x] != 'F' {
+			return false
+		}
+	}
+
+	return true
+}
+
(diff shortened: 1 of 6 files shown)


stackdump:
I also renamed one of the options in ./test.sh  changed from 'long' to 'full' 
+ added a usage helper

wsapi/debugapi.go
```diff
@@ -427,15 +427,16 @@ func runCmd(cmd string) {
 
 	return
 }
+
 func HandleSimControl(state interfaces.IState, params interface{}) (interface{}, *primitives.JSONError) {
-	droprate := new(GetCommands)
-	err := MapToObject(params, droprate)
+	cmdLines := new(GetCommands)
```
stackdump: renamed b/c it didn't look like we are limited to droprate commands here

testHelper/simulation.go
```diff
@@ -129,105 +130,175 @@ func StartSim(GivenNodes string, UserAddedOptions map[string]string) *state.Stat
...
-	if len(creatingNodes) > nodes {
-		t.Fatalf("Should have allocated %d nodes", len(creatingNodes))
-	}
-	WaitForMinute(state0, 1)
+func promoteNodes(creatingNodes string) int {
+
 	for i, c := range []byte(creatingNodes) {
 		fmt.Println("it:", i, c)
 		switch c {
-		case 'L', 'l':
```
stackdump: I don't think we use lower anywhere - moving to try to standardize on caps

wsapi/debugapi.go
factom-clay: good change

testHelper/simulation.go
factom-clay: What's the benefit to being case sensitive?

PR #815, opened 2019-07-25
Added of date for data commit entry to endpoint "ack"

ilzheev:
Our core developer added `transactiondate` & `transactiondatestring` into factomd API `ack` response.

So, `ack` of entry now contains the timestamp of entry creation (transaction).
Looks like this timestamp existed in `entry-ack`, but was lost during refactoring from `entry-ack` → `ack`, now we return it back.

I have tested it and it works correctly.

### DIFF
--- a/wsapi/ack.go
+++ b/wsapi/ack.go
@@ -186,6 +186,16 @@ func handleAckByEntryHash(hash interfaces.IHash, state interfaces.IState) (inter
 			answer.CommitTxID = txid.String()
 			answer.CommitData.Status = constants.AckStatusString(constants.AckStatusDBlockConfirmed)
 		}
+		_, _, txTime, _, err := state.GetSpecificACKStatus(txid)
+		if err != nil {
+			return nil, NewInternalError()
+		}
+		if txTime != nil {
+			answer.CommitData.TransactionDate = txTime.GetTimeMilli()
+			if txTime.GetTimeMilli() > 0 {
+				answer.CommitData.TransactionDateString = txTime.String()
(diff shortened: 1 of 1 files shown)


PaulBernier:
the `datestring` is redundant information with the timestamp and less flexible. It should be the responsibility of the client to format the time to a string as it pleases him/her. Basically I think the `datestring` should be simply dropped.

ilzheev:
> the `datestring` is redundant information with the timestamp and less flexible. It should be the responsibility of the client to format the time to a string as it pleases him/her. Basically I think the `datestring` should be simply dropped.

I agree with that, but from the consistency side it should be there.
There is `datestring` for `entrydata`:
```json
…
"entrydata": {
"blockdate":1564044300,
"blockdatestring":"2019-07-25 12:45:00",
…
```

So, Michael added `transactiondate` & `transactiondatestring`, so both `commitdata` & `entrydata` will be consistent.

ilzheev:
We are preparing a proposal for refactoring of `ack` call, the drop of `datestring`'s should be discussed there IMO.

PaulBernier:
Ah, sure, I didn't realize that `datestring` was not your addition.

ilzheev:
@stackdump @Emyrk  Hey, I'm testing `v6.4.2-rc2-alpine` and looks like this PR is not included there. Can we include it into final Xuan release?

stackdump:
added another pr: https://github.com/FactomProject/factomd/pull/885  to pull this commit into rollup 3

PR #825, opened 2019-08-03
Xuan factable changes try 3

factablesolutions:
Code modifications amidst comments.

### DIFF
--- a/common/primitives/binary.go
+++ b/common/primitives/binary.go
@@ -14,6 +14,7 @@ import (
 	"github.com/FactomProject/factomd/common/primitives/random"
 )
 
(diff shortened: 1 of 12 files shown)


stackdump:
Added a backlog item to investigate why docker build is failing 

Everything looks good to me

factablesolutions:
Thanks Matt! I also wasn't sure about the docker failure, but I got that with my previous checkin which was 99% comments and maybe a receiver name change or two. So I just figured it wasn't associated with my changes, but let me know if for some reason its an issue on my side.

stackdump:
Oh it's def some config issue on our end - Circle Lets use enable builds on non-FactomProject accounts (which we recently enabled)

I think a side effect is this build failure since we have a stored credential for pushing docker builds.

PR #832, opened 2019-08-15
Fd 1150 live api

sanderPostma:
Initial working implementation for review.


events/eventservices/eventsService.go
```diff
@@ -0,0 +1,221 @@
...
+
+	factomEvent, err := MapToFactomEvent(event)
+	if err != nil {
+		return fmt.Errorf("failed to map to factom event: %v\n", err)
+	}
+	factomEvent.IdentityChainID = &eventmessages.Hash{
+		HashValue: esi.owningState.GetIdentityChainID().Bytes(),
+	}
+	select {
+	case esi.eventsOutQueue <- factomEvent:
```
Emyrk: This is dropping the most recent messages if the queue is full, so the queue grows stale. If we could drop off of the front when the queue is full, that would be ideal. The implementation get's a bit tricky since different threads can add/remove from the queue, so I think we might just have to live with this.

I'd like to suggest adding a counter here that counts the number of messages dropped from back pressure. We could make this counter accessible via a prometheus counter, and via a `/v1/livefeed` endpoint for livefeed specific api calls. This would let a client identify if they are potentially losing information.

sanderPostma:
I've revisited the protobuf structures a bit. Can you please check the files in here if this is what you envision?
https://github.com/bi-foundation/factomd/tree/FD-1150_live_api/events/eventmessages
mainly here https://github.com/bi-foundation/factomd/blob/FD-1150_live_api/events/eventmessages/factomEvents.proto
- Renamed AnchorEvent to DirBlockConsensus because anchored was open to interpretation. If you know a better name please let me know.
- EventSource became StreamSource to indicate if it's a live or replay stream
- ChainRegistration & EntryRegistration will be sent as soon as they are registered and won't be sent again for the state changes. (Only the directory block will contain all entries again because we only get new chain & entry registrations from the current factomd node and the dirblock contains the entries from all nodes who added entries to the process list.)
- StateChanged will only send the entry/chain hash along with the new state.

sanderPostma:
### Some notes and questions concerning the proposed changes
The initial requirement was to keep things very simple on the factomd side and give the second layer more control over what's  Some of the describe features means that we're going to pull in more functionality to the first layer and at the same time require more logic in the second layer.

#### Send content once: 
This will reduce load to the network (although this will usually be to a loopback interface and serialization is quite fast) but this will require the second layer to either remember or request the chain and entries again to be able to track to which entity a state change applies to.

#### Some filter options:
We have already built that into the second layer (using GraphQL) and the chain & entry lifecycles only apply to the local node. Should we also break up the directory block, filter for only certain transaction addresses and filter out things like admin blocks? You want to control these filter settings using a REST endpoint? When the filter requirements don't change you can keep them in the config file plain and simple, but if they need to be configured dynamically I am going to have more questions like what to do with these settings when the node shuts down.

sanderPostma:
Since different use cases will require different behaviour I've added two config switches to have some control over filtering.
ResendRegistrationsOnStateChange bool
-> if you don't want to or can't remember/track entries in the second layer you can set this to true. Otherwise (by default )it will send the chain & entry registrations only once followed by only StateChange events.

ContentFilterMode           contentfiltermode.ContentFilterMode
	SendNever                ContentFilterMode = 0
	SendOnRegistration ContentFilterMode = 1
	SendAlways              ContentFilterMode = 2
	Unknown                  ContentFilterMode = -1
--> Self explanatory. The current version has a "EntryContentRegistration" again which is basically the RevealEntry and will only be send when teh filtermode > SendNever
Also the directory block will only include the content (& ext id's) in SendAlways mode.

sanderPostma:
De renamed some fields to be more friendly to the API users.
message FactomEvent {
    StreamSource streamSource = 1;
    Hash identityChainID = 2;
    oneof value {
        ChainRegistration chainRegistration = 3;
        EntryRegistration entryRegistration = 4;
        EntryContentRegistration entryContentRegistration = 5;
        StateChange stateChange = 6;
        BlockCommit blockCommit = 7;
        ProcessMessage processMessage = 8;
        NodeMessage nodeMessage = 9;
    }
}

Maybe I should also rename ProcessMessage and NodeMessage to something like BlockchainStateMessage and ServerMessage. (I'm trying to see this from the perspective of a developer who'be never had to do anything with Factom or Blockchain until today.)

sanderPostma:
Prometheus counters were added

`		eventServiceInstance.droppedFromQueueCounter = prometheus.NewCounter(prometheus.CounterOpts{
			Name: "factomd_livefeed_dropped_from_queue_counter",
			Help: "Number of times we dropped events due of a full the event queue",
		})
		eventServiceInstance.notSentCounter = prometheus.NewCounter(prometheus.CounterOpts{
			Name: "factomd_livefeed_not_send_counter",
			Help: "Number of times we couldn't send out an event",
		})
`

sanderPostma:
For the replay functionality, is the REST endpoint crucial for that? Or could that be broken up in two phases? Like initially using the command line and later using a REST endpoint.

lvandenbrink:
I published a simple application that subscribes on the live-feed-api and stores the event on disk: https://github.com/bi-foundation/live-feed-subscriber To run the application factomd should be started with the config "EnableLiveFeedAPI = true", and the live-feed-api from the develop branch.

sanderPostma:
Besides the "submit replay job" functionality this should be it

engine/factomParams.go
```diff
@@ -93,6 +93,17 @@ func init() {
 	flag.StringVar(&p.ControlPanelSetting, "controlpanelsetting", "", "Can set to 'disabled', 'readonly', or 'readwrite' to overwrite config file")
 	flag.BoolVar(&p.FullHashesLog, "fullhasheslog", false, "true create a log of all unique hashes seen during processing")
 	flag.BoolVar(&p.ReparseAnchorChains, "reparseanchorchains", false, "If true, reparse bitcoin and ethereum anchor chains in the database")
+
+	// Live feed API params
+	flag.BoolVar(&p.EnableLiveFeedAPI, "enablelivefeedapi", false, "Enable life feed events service; default false")
+	flag.StringVar(&p.EventReceiverProtocol, "eventreceiverprotocol", "", "Transport protocol for the events receiver; default udp")
+	flag.StringVar(&p.EventReceiverAddress, "eventreceiveraddress", "", "Address for the events receiver; default 127.0.0.1")
+	flag.IntVar(&p.EventReceiverPort, "eventreceiverport", 0, "Port for the events receiver; default 8040")
+	flag.StringVar(&p.OutputFormat, "OutputFormat", "", "Event format for the events receiver, protobuf|json; default protobuf")
```
hannemennah: This should be lowercase, to be consistent with the other flag options.

events/README.md
```diff
@@ -0,0 +1,54 @@
...
+|  MuteReplayDuringStartup          | At startup factomd can replay all the events that were stored since that last fastboot snapshot. Use this property to turn that on/off.   | true &#124; false |
+|  ResendRegistrationsOnStateChange | It’s possible to choose whether the chain and entry commit registrations should only be sent once, followed by state change events vs resending them for every state change. The first option reduces overhead & network traffic, but requires the implementer to track which state changes belong to which chain or entry.| true &#124; false |
+|  ContentFilterMode                | This option will determine whether the external ID’s and content will be included in the event stream. There are three level settings for this. Please note that the combination of ResendRegistrationsOnStateChange = true and ContentFilterMode=Always will resend all dataon every state change. The maximum content size per entry is only 10KB, however with a large number of transactions per second this may add up to an undesirable amount of data. | Always &#124; OnRegistration &#124; Never |
```
hannemennah: README documentation calls this "Always" but it should be "SendAlways" instead, because that's what the actual parameter is implemented as. (Alternatively, keep the documentation as "Always" and update instances of "SendAlways" to "Always"... either way works. Just need to make sure they match.)

hannemennah:
Great work on this, guys. I've been testing and working with it for the past few days and I'm really impressed with the job you've done.

The biggest issue I've encountered is that dbstate information doesn't propagate through the LiveFeed _when the feeding node is blockwise in-sync with the leaders of the network_. In other words, when the dbState has `isNew` set as `true`, no dbstate information reaches the LiveFeed.

It makes sense why this is happening; the dbstate is built from process list data, and so no dbstate message ever has to be sent (either from the other nodes on the network, or from the local node doing the building). Since no dbstate message is sent or received, the LiveFeed never picks one up, and the dbstate is "invisible" from its perspective. However, a lot of the value we are hoping to get out of the LiveFeed API is the ability to monitor the blockchain in real-time (including dblock, ablock, fblock, etc. info) and without some sort of mechanism to capture this data through the LiveFeed, we miss out on a large portion of that ability.

There are a few ways to approach this; a dbstate message could be generated and processed locally simply for the sake of the LiveFeed, or another hook (or more than one) could be inserted into the code to trigger LiveFeed events even without the relevant messages involved. I don't have any strong feelings or suggestions as to how it should be implemented, just hoping to see it added before everything gets merged.

If you need replication steps, the easiest way to replicate is by running a 2-node LOCAL or CUSTOM network (one Leader and one Follower)... as the Follower catches up to the Leader's height, it will receive and process dbstates successfully (and these are captured by the LiveFeed), but once it has synced to the tip and is following along at the ProcessList level, no more dbstates (or directoryBlockCommits) come through. Instead, new minute information and new block information continue to be emitted (along with the info associated with any entry/transactions you are making locally).

events/README.md
sanderPostma: I still had some changes for that which were not pushed yet. When you pull again ContentFilterMode will be called AllowContent. The combination filter and always/never can be confusing. (is it always filtering or always sending)

sanderPostma:
I have moved the emitter point to dbStateManager.UpdateState() and slightly altered the mapper to deal with the DBState directly (through interface IDBState)
Now it's emitting dirblock events also after it's synced up. I also ran into the issue that these events ran through dbStateManager.UpdateState twice and therefore were emitted twice. It appears that after the node is synced up or maybe slightly lagging behind, there are still coming in dbstate messages and these blocks get double- saved to the database. I've added a condition for this to make sure it's only emitted when it's saved for the first time.

I've also seen inconsistencies with new block / minute which are also message driven and behave kind of unpredictable. I'll be looking into that.

sanderPostma:
I've done some work getting the unit tests to work in CircleCI. They are passing now.

sambarnes:
**Summary of feedback**:
-  no Entry Blocks included in the directory block commits
- I'm seeing `directoryBlockCommit.directoryBlock.timestamp` is always showing as a blank object `{}`
- in the directory block structure, the hash labels are named incorrectly for admin blocks and entry credit blocks, they should be lookupHash and headerHash respectively 
- probably should also include the keymrs in the root level of all of the blocks as well, so that you don't force users to do that calculation themselves in a language that likely doesn't have a factom library (or at the least do it for directory blocks)
- I see that factoid blocks are inconsistent compared to the rest of them. FBlock is the only one not split into `block.header` and `block.entries`/`block.transactions`
- in your factoid transactions, they're missing RCD datastuctures:
```
...
"redeemConditionDataStructures": [
          {
            "Rcd": null
          },
          {
            "Rcd": null
          }
],
...
```
- also in the factoid transaction datastructure, the signatureBlock seems to be in a weird fomat, could you explain why this is? I would think that it would be split into a signature and a public key (since its named `signature` and not `fullSignature`). And also peculiar that a "signature" wouldnt be an array with a single string value
```
"signatureBlocks": [
          {
            "signature": [
              "Q39IfsZMkutZXjH0wcBgJEkWNJY4FMywp5rdtIH3yyqMiR4q8vKqq8sNCgoBONEYnA8cbUrdxpK8/BWMLGD5Bg=="
            ]
          },
          {
            "signature": [
              "7i+ZLbOHP3ie2aCdOpW+xwv3nAjROo+AUFkn5J62TTTTAht1vLdsctF6KQqlHKPT4KpsDQtka+CACW1evx1jBw=="
            ]
          }
]
```
- i see that fct transactions have a timestamp but no minute number. can you detail the reasons behind this decision? I think minute markers are useful in addition to timestamp so that a consumer can construct the factoid block body and marshal it if they wanted to construct the full factoid block object and validate it. As is, a user cannot tell where in the fblock the tx fell
- the `timestamp` elements are always split like `{"seconds": 1, "nanos": 6180000}` but it would just save space and make more sense if it were a single integer timestamp as they appear in the underlying datastructure
- factoid blocks don't have  body size and messagee count included like admin blocks do

A lot of my feedback can be boiled down to just modelling the JSON datastructure after the actual underlying things they represent and including all of those fields: https://github.com/FactomProject/FactomDocs/blob/master/factomDataStructureDetails.md since those are the blockchain primitives that we are working with and should be mirroring in our code

events/README.md
```diff
@@ -0,0 +1,45 @@
...
+|  EventReplayDuringStartup         | At startup factomd can replay all the events that were stored since that last fastboot snapshot. Use this property to turn that on/off.   | true &#124; false |
+
+The same properties can be overridden by command line parameters which are the same as above but lowercase.
+The retry mechanism of the first layer is pretty strict. When a receiver is down or for some reason unresponsive it will retry to connect 3 times. If a receiver is not up by then, it will keep retrying to restore the connection every 5 minutes, but in the meantime it will start dropping the events until the receiver is back up. For mission critical use-cases there are prometheus counters in place:
+* **factomd_livefeed_not_send_counter** - the number of events that came out of the channel, but could delivered to the receiver.
```
sambarnes: Looks to be a typo in this line

sanderPostma:
> A lot of my feedback can be boiled down to just modelling the JSON datastructure after the actual underlying things they represent and including all of those fields
First of all, the primary design & implementation was of the live feed is protobuf, json we just did as an extra because it was an very easy feature to implement.
But we'll unwrap what can be unwrapped without breaking the profobuf/graphql implementation.  Some of the issues you described are mirrored from the backend struct so we need to decide if/what we can improve. For some points I will take the discussion to Discord.

sanderPostma:
> I'm seeing directoryBlockCommit.directoryBlock.timestamp is always showing as a blank object {}

Fixed by Laurens

sanderPostma:
> in the directory block structure, the hash labels are named incorrectly for admin blocks and entry credit blocks, they should be lookupHash and headerHash respectively.

Most of these names were derived from how these were called in the code struct or interface function.  Which exact  field names should be called lookupHash and headerHash?
Also the admin block's previousBackRefHash is a bit inconsistent, this should maybe be previousAdminBlockHash. I'll ask on Discord.

sanderPostma:
> probably should also include the keymrs in the root level of all of the blocks as well, so that you don't force users to do that calculation themselves in a language that likely doesn't have a factom library (or at the least do it for directory blocks)

Currently only DirectoryBlockEntry has it. We'll add it for other entities that have a GetKeyMR() function.

sanderPostma:
> I see that factoid blocks are inconsistent compared to the rest of them. FBlock is the only one not split into block.header and block.entries/block.transactions
This is currently mirroring the underlying structure, but we can make it more consistent by either adding a header to the factoid block, or merging the header fields from the other blocks to the root structure. Will ask on discord.

sanderPostma:
> in your factoid transactions, they're missing RCD datastuctures

RCD population was fixed by Laurens
We can also unwrap them, but the underlying struct shows a possible "RCD2" option. If RCD2 is not going to be a thing anymore, we can unwrap it, but when RCD2 is future feature we'll need this extra level to switch between RCD1 & RCD2.

> RCD should be all caps or lowercase since its a single acronym
It appears that no lower-case json annotation is generated for oneOf objects and the R is made upper-case because it's a public field. 
If we remove RCD support this issue will be gone along with it.

sanderPostma:
> also in the factoid transaction datastructure, the signatureBlock seems to be in a weird fomat, could you explain why this is? Peculiar that a "signature" would be an array with a single string value. It probably makes more sense to include these in the rcd section like the underlying data structure does (something like {"rcd": {...}, "signature": "..."} would be better in a single section)
"signatureBlocks": [
         {
           "signature": [
             "Q39IfsZMkutZXjH0wcBgJEkWNJY4FMywp5rdtIH3yyqMiR4q8vKqq8sNCgoBONEYnA8cbUrdxpK8/BWMLGD5Bg=="
           ]
         },
         {
           "signature": [
             "7i+ZLbOHP3ie2aCdOpW+xwv3nAjROo+AUFkn5J62TTTTAht1vLdsctF6KQqlHKPT4KpsDQtka+CACW1evx1jBw=="
           ]
         }
]

In protobuf this is less of an issue than with json because the structures are precompiled. We can see if we can unwrap some more structures, but in some cases it's better for object marshallers to work with objects. Also with protobuf oneOf objects we always need to wrap the fields in objects because they can't contain primitives. As mentioned above, we'll unwrap what we can.

sambarnes:
> > I see that factoid blocks are inconsistent compared to the rest of them. FBlock is the only one not split into block.header and block.entries/block.transactions
> This is currently mirroring the underlying structure, but we can make it more consistent by either adding a header to the factoid block, or merging the header fields from the other blocks to the root structure. Will ask on discord.

The underlying structure is that all blocks have headers, including factoid blocks https://github.com/FactomProject/FactomDocs/blob/master/factomDataStructureDetails.md#factoid-block But it is missing in the your fblock implementation. I think they should all have headers and bodies, rather than flatten all other blocks.


> > in the directory block structure, the hash labels are named incorrectly for admin blocks and entry credit blocks, they should be lookupHash and headerHash respectively.
> 
> Most of these names were derived from how these were called in the code struct or interface function. Which exact field names should be called lookupHash and headerHash?
> Also the admin block's previousBackRefHash is a bit inconsistent, this should maybe be previousAdminBlockHash. I'll ask on Discord.

In your dblock entries, inside a dblock, inside a dblock commit. The entries all have "chainID" and "keyMerkleRoot" in your implementation. This is incorrect for the Admin Block and Entry Credit block entries there though. The values you used are the ABlock.lookupHash (not backreference hash) and the ECBlock.headerHash, and should be named as such. See the body of this section: https://github.com/FactomProject/FactomDocs/blob/master/factomDataStructureDetails.md#directory-block a/ec blocks aren't represented keymrs


> > in your factoid transactions, they're missing RCD datastuctures
> 
> RCD population was fixed by Laurens
> We can also unwrap them, but the underlying struct shows a possible "RCD2" option. If RCD2 is not going to be a thing anymore, we can unwrap it, but when RCD2 is future feature we'll need this extra level to switch between RCD1 & RCD2.

That is a good point. I think we should keep these rcd structures wrapped then, to allow the distinction between types.

sanderPostma:
>  see that fct transactions have a timestamp but no minute number. can you detail the reasons behind this decision?
We use the timestamp from google: https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/Timestamp

sanderPostma:
> the timestamp elements are always split like {"seconds": 1, "nanos": 6180000} but it would just save space and make more sense if it were a single integer timestamp as they appear in the underlying datastructure I still don't agree with the use of this really and haven't seen any rationale behind its use

This is the standard way a Google/golang timestamp works. Nanosecond precision without overflowing an integer.

sanderPostma:
> ensure that the timestamps match the underlying timestamp for the raw block or block element (e.g. dblocks have a timestamp in minutes rather than milliseconds)

This is handled by existing code in directoryBlockHeader.go
`func (h *DBlockHeader) GetTimestamp() interfaces.Timestamp {
    return primitives.NewTimestampFromMinutes(h.Timestamp)
}`

sambarnes:
> > see that fct transactions have a timestamp but no minute number. can you detail the reasons behind this decision?
>
> We use the timestamp from google: https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/Timestamp

I realize that. However there are two notions of time in a factoid transaction. The timestamp on the tx, and the minute markers between txs in the block. Minute markers are needed to be able to marshal a factoid block from the live feed output

sanderPostma:
> factoid blocks don't have body size and message count included like admin blocks do

These do not exist in the underlying FBlock struct. We could count them and add them, if they are useful. Or protobuf implementation doesn't use those.

sanderPostma:
> admin block entries are wrapped unnecessarily and do not need a "AdminBlockEntry" elements since its already in an entries array inside an admin block

same as answer before

sanderPostma:
> admin block entries should include the admin id integer for consumers to easily know what the type is

I've omitted this field because the protobuf marshaller already has its own type id and for that it would be duplicate. I can plug the original admin id field which matches the documentation. (It's just a byte)

sambarnes:
> > A lot of my feedback can be boiled down to just modelling the JSON datastructure after the actual underlying things they represent and including all of those fields
> 
> First of all, the primary design & implementation was of the live feed is protobuf, json we just did as an extra because it was an very easy feature to implement.
> But we'll unwrap what can be unwrapped without breaking the profobuf/graphql implementation. Some of the issues you described are mirrored from the backend struct so we need to decide if/what we can improve. For some points I will take the discussion to Discord.

Like I've mentioned before, I think it is unwise for a deliberately separate output format to be influenced/determined by the other output format that the user specifically chose not to use. Internally, I would expect the underlying go object to be passed around until the very end where it is either:
- sent using  OriginalGoStruct.MarshalJSON()
- send using OriginalGoStruct mapped to NewProtobufGoStruct

sanderPostma:
> also, minute number in EC Block entries is redundant and can just be {"minuteNumber": 1}:

In this case it will break our GraphQL implementation if we unwrap it, the first minuteNumber is the container object, the second is the uint32 value (same goes for ServerIndexNumber) We can't generate "oneof" entities with primitives, they always need to be objects.

sambarnes:
> > factoid blocks don't have body size and message count included like admin blocks do
> 
> These do not exist in the underlying FBlock struct. We could count them and add them, if they are useful. Or protobuf implementation doesn't use those.

See the bottom of the  header, they are part of the struct
https://github.com/FactomProject/FactomDocs/blob/master/factomDataStructureDetails.md#factoid-block

events/README.md
lvandenbrink: I corrected the typo

lvandenbrink: I corrected the text

sanderPostma:
> The underlying structure is that all blocks have headers, including factoid blocks https://github.com/FactomProject/FactomDocs/blob/master/factomDataStructureDetails.md#factoid-block But it is missing in the your fblock implementation

That's what the documentation says, but the actual implementation is like this:
image
The header fields are in the same struct, opposed to as it is like in the other block types which have the header fields in a separate header object like DirectoryBlock -> DBlockHeader.

sambarnes:
> > The underlying structure is that all blocks have headers, including factoid blocks https://github.com/FactomProject/FactomDocs/blob/master/factomDataStructureDetails.md#factoid-block But it is missing in the your fblock implementation
> 
> That's what the documentation says, but the actual implementation is like this:
> image
> The header fields are in the same struct, opposed to as it is like in the other block types which have the header fields in a separate header object like DirectoryBlock -> DBlockHeader.

You and I both know there are plenty of decisions in factomd that should be reworked and didn't follow spec exactly haha ;) This is a case where if we're adding new code, we should make the new code follow official spec rather than propagating the mistakes of the past IMO

sanderPostma:
> Minute markers are needed to be able to marshal a factoid block from the live feed output

Do you mean the minute marker with in block on which the transaction took place? I don't see this level of detail in the ITransaction interface. Is that even recorded?

sambarnes:
> > Minute markers are needed to be able to marshal a factoid block from the live feed output
> 
> Do you mean the minute marker with in block on which the transaction took place? I don't see this level of detail in the ITransaction interface. Is that even recorded?

Yep the minute markers between transactions are encoded as ten single "0x00" bytes: 
```
The Factoid Block consists of a header and a body. The body is composed of serialized Factoid transactions with minute markers distributed throughout the body. The minute markers consist of a single byte 0x00. There are 10 of them, each being placed at a minute boundary, and the 10th marker being the last item in the block. Factoid transactions begin with a version number above zero, which is how they can be differentiated from transactions. The minute markers are included in both the body and ledger Merkle roots.
```

IIRC it was implemented in a somewhat convoluted way in the actual FBlock struct in go, rather than doing a map from integer minutes to slices of transactions within that minute

sanderPostma:
> Internally, I would expect the underlying go object to be passed around until the very end where it is either:
sent using OriginalGoStruct.MarshalJSON()
send using OriginalGoStruct mapped to NewProtobufGoStruct

Not impossible but will take some doing. Also we will lose the ability to change abbreviated names and correct inconsistencies like introducing a separate header struct for fblock.

sambarnes:
> > Internally, I would expect the underlying go object to be passed around until the very end where it is either:
> > sent using OriginalGoStruct.MarshalJSON()
> > send using OriginalGoStruct mapped to NewProtobufGoStruct
> 
> Not impossible but will take some doing. Also we will lose the ability to change abbreviated names and correct inconsistencies like introducing a separate header struct for fblock.

IIUC we would not lose the ability, we could just make a new json marshalling function (UnmarshalJSONV2 perhaps) for the structures and make sure it mimicks the documentation as much as possible. This could potentially serve as the basis for the v3 RPC API if designed consistently enough across types (again, something we both know needs fixing in factomd haha)

Then when the rewrite comes along, those V2 functions can just become the official JSON marshal/unmarshallers

sambarnes:
> The factomDataStructureDetails document specifically describes how the messages are sent over the P2P network, I see a part of the fields are inserted on the fly in .MarshalBinary() and are not present in the structure that is saved to the database (where I intercept the event)
> Some of these fields are also only interesting for the P2P protocol (ie. transaction count/body size). But I can look at the MarshalBinary() code to see what's going on with the minute markers.

No that document isn't really about P2P. Sure, that is part of how they're sent around (in a specific marshalled binary format) because that is the smallest footprint possible. The P2P protocol is more abstracted away than these structures, and wraps these binary payloads in different message types. But that isn't described in that doc.

The document describes **the** core data structures of the blockchain, and they are stored in the database in that exact format, byte for byte. They are the basis of the blockchains auditability (arguably factoms core value prop). If you want to validate Merkle Roots/Entry Hashes/etc etc, you need to be able to marshal the data in that exact format. Otherwise you cannot form proofs of the data. Hence why I think everything a user will need to create proofs should be included.

stackdump:
Going to pull this in & sync up w/ A4 pre-release version for more internal testing

if there are changes left out of this PR we can do another merge

PR #840, opened 2019-08-22
Fd 1161 xuan rollup3 test sync

PaulSnow:


PaulSnow:
The useful code is already in FD-1161_rollup3_test

PR #847, opened 2019-08-28
let message filter change via heartbeat

stackdump:
port heartbeat filter tweaks that were coded against bond


stackdump:
replaced this PR

PR #856, opened 2019-09-03
replace a message if output queue is full during a wait period

stackdump:
don't send out messages until we are outside of a wait period

Also replace messages if queue ends up being full during this situation


stackdump:
I'll re-open against proper branch

PR #860, opened 2019-09-03
Removed remainig part of NO_SORT activation for election.

factom-clay:


stackdump:
failure is from `./wsapi/...` module - we can tackle this after merge

PR #871, opened 2019-09-05
Fd 1132 debug rpc enhancement

stackdump:
Targeting Xuan_rollup3 for Review - should change base to A4 when it's up-to-date


stackdump:
NOTE: we will still need to update The Docs for Debug api (which is not used when connecting to mainnet)

PR #872, opened 2019-09-06
Fd 1189 simtest lagging audit election

stackdump:
I think this is what was asked for - adding @factom-clay to review

Upon futher reflection - I'm unsure what we are actually testing for.
Is the thing we are testing for that the network will rebound within a certain time period?


stackdump:
I think Clay had a different version of this that works

PR #873, opened 2019-09-06
Fixed expiration from holding to happen once per block in minute 2. F…

factom-clay:
…ixed holding accounting bugs.

### DIFF
--- a/state/stateConsensus.go
+++ b/state/stateConsensus.go
@@ -561,6 +561,53 @@ func CheckDBKeyMR(s *State, ht uint32, hash string) error {
 // Consensus Methods
 //***************************************************************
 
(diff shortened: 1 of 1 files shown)


state/stateConsensus.go
```diff
@@ -728,8 +727,6 @@ processholdinglist:
 		switch validToExecute {
 		case -1:
 			s.LogMessage("executeMsg", "invalid from holding", v)
-			TotalHoldingQueueOutputs.Inc()
```
stackdump: Do we care about losing all these metrics?

stackdump: Clay let me know that there was a bug here - removing these counters is desired.

PR #876, opened 2019-09-10
FD-1189 simtest lagging audit election

stackdump:
TODO: still want to look into failure of second case


stackdump:
Will cherry pick this test into A4 if we still find it useful.

PR #881, opened 2019-09-19
add threaded response for missing data

stackdump:


engine/NetworkProcessorNet.go
```diff
@@ -103,7 +104,7 @@ func Peers(fnode *FactomNode) {
 			case constants.MISSING_DATA:
 				if !fnode.State.DBFinished {
 					return true
-				} else if fnode.State.InMsgQueue().Length() > constants.INMSGQUEUE_HIGH {
+				} else if fnode.State.DataMsgQueue().Length() > constants.INMSGQUEUE_HIGH { // REVIEW: should this limit be altered?
```
stackdump: @factom-clay do you think this same limit is OK to use here?

engine/NetworkProcessorNet.go
```diff
@@ -389,6 +393,12 @@ func Q2(fnode *FactomNode, source string, msg interfaces.IMsg) {
 	fnode.State.InMsgQueue2().Enqueue(msg)
 }
 
+func DataQ(fnode *FactomNode, source string, msg interfaces.IMsg) {
+	fnode.State.LogMessage("NetworkInputs", source+", enqueue_data", msg)
```
factom-clay: Not adding it to network in so don't log it in networkinputs.

engine/NetworkProcessorNet.go
```diff
@@ -389,6 +393,12 @@ func Q2(fnode *FactomNode, source string, msg interfaces.IMsg) {
...
+	fnode.State.LogMessage("DataQueue", source+", enqueue_data", msg)
```
factom-clay: just enqueue the log names says where it is being enqueue to.

engine/NetworkProcessorNet.go
```diff
@@ -534,3 +544,13 @@ func InvalidOutputs(fnode *FactomNode) {
 		// }
 	}
 }
+
+// Handle requests for missing data
+func MissingData(fnode *FactomNode) {
+	for {
+		time.Sleep(1 * time.Millisecond)
```
factom-clay: Why sleep?

engine/NetworkProcessorNet.go
```diff
@@ -534,3 +544,13 @@ func InvalidOutputs(fnode *FactomNode) {
...
+		msg := fnode.State.DataMsgQueue().BlockingDequeue()
+		fnode.State.LogMessage("DataQueue", "dequeue_data", msg)
```
factom-clay: just dequeue. file name says where. log analysis tools match enqueue/dequeue messages.

state/state.go
```diff
@@ -179,6 +179,7 @@ type State struct {
 	apiQueue               APIMSGQueue
 	ackQueue               chan interfaces.IMsg
 	msgQueue               chan interfaces.IMsg
+	dataQueue              InMsgMSGQueue
```
factom-clay: Do you need the fancy stuff a InMsgQueue has or would a channel have been just fine?

state/state.go
```diff
@@ -1009,6 +1010,7 @@ func (s *State) Init() {
 	s.msgQueue = make(chan interfaces.IMsg, 50)                             //queue of Follower messages
 	s.prioritizedMsgQueue = make(chan interfaces.IMsg, 50)                  //a prioritized queue of Follower messages (from mmr.go)
 	s.MissingEntries = make(chan *MissingEntry, constants.INMSGQUEUE_HIGH)  //Entries I discover are missing from the database
+	s.dataQueue = NewInMsgQueue(constants.INMSGQUEUE_HIGH)                  //incoming requests for missing data
```
factom-clay: INMSGQUEUE_HIGH is a big size? Processing should be pretty fast. Do you have reason to thing it needs to be big ? The channel for missing_messages is like 20 messages deep.

engine/NetworkProcessorNet.go
factom-clay: In general, unless it's required to prevent deadlock, I'm against dropping messages.

state/state.go
stackdump: no reason - I can set it to something reasonable like 20

factom-clay:
Did you do any experimentation to see if it affected performance?

state/state.go
stackdump: channel would have worked

engine/NetworkProcessorNet.go
stackdump: no reason really - copy behavior from other goroutines

stackdump: no good reason - copy from other loops

engine/NetworkProcessorNet.go
```diff
@@ -389,6 +390,12 @@ func Q2(fnode *FactomNode, source string, msg interfaces.IMsg) {
 	fnode.State.InMsgQueue2().Enqueue(msg)
 }
 
+func DataQ(fnode *FactomNode, source string, msg interfaces.IMsg) {
+	q := fnode.State.DataMsgQueue()
+	fnode.State.LogMessage("DataQueue", fmt.Sprintf(source+", enqueue %v", len(q)), msg)
+	q <- msg
```
Emyrk: Do we want to drop requests if this backs up?

I don't see a deadlock every happening, but a ton of requests could bog down a node. I'd rather my node drop a request than block.

stackdump: I don't think this is configured w/ a max size

state/state.go
stackdump: @Emyrk ^ I probably have a bug here - this is a very large buffer

stackdump: ok not a bug exactly - maybe should be `make(chan interfaces.IMsg, constants.INMSGQUEUE_HIGH) ` instead

PR #884, opened 2019-09-25
FD-1207 Simplify timer, address syncopation

PaulSnow:
Remove old logging, sync to the start of the minute with each pass of the timer, remove complexity (that might be subject to slippage) from the syncing to minute boundaries.


engine/timer.go
```diff
@@ -28,9 +29,15 @@ func Timer(stateI interfaces.IState) {
...
-			s.TickerQueue() <- -1 // -1 indicated this is real minute cadence
+			if s.LastSyncTime.Nanoseconds() > tenthPeriod*9/10 && s.Leader {
+				s.LastSyncTime = 0
+				_, _ = os.Stderr.WriteString(fmt.Sprintf("Slipping a minute %10s %d  dbht %5d\n", s.FactomNodeName, s.CurrentMinute, s.LLeaderHeight))
 
-			s.LogPrintf("ticker", "Tick! %d, tenthPeriod=%s", i, time.Duration(tenthPeriod))
+			} else {
+				s.TickerQueue() <- -1 // -1 indicated this is real minute cadence
+
+				s.LogPrintf("ticker", "Tick! %d, tenthPeriod=%s", i, time.Duration(tenthPeriod))
```
factom-clay: This behavior eliminates the unused relationship between i and current minute.  Maybe drop the "for i := 0; i < 10; i++" loop?

engine/timer.go
```diff
@@ -28,9 +29,15 @@ func Timer(stateI interfaces.IState) {
 			// Leaders running with slightly different minutes in test environments.
 			time.Sleep(time.Duration(s.GetTimeOffset().GetTimeMilli()) * time.Millisecond)
 
-			s.TickerQueue() <- -1 // -1 indicated this is real minute cadence
+			if s.LastSyncTime.Nanoseconds() > tenthPeriod*9/10 && s.Leader {
+				s.LastSyncTime = 0
+				_, _ = os.Stderr.WriteString(fmt.Sprintf("Slipping a minute %10s %d  dbht %5d\n", s.FactomNodeName, s.CurrentMinute, s.LLeaderHeight))
```
factom-clay: Log this?

engine/loadcreate.go
```diff
@@ -83,7 +83,7 @@ func (lg *LoadGenerator) Run() {
 			s := fnodes[wsapiNode].State
 			s.APIQueue().Enqueue(c)
 			s.APIQueue().Enqueue(r)
-			time.Sleep(time.Duration(800/top) * time.Millisecond) // spread the load out over 800ms + overhead
+			//time.Sleep(time.Duration(800/top) * time.Millisecond) // spread the load out over 800ms + overhead
```
factom-clay: At what load did calling sleep here plus creating the messages cause the load generation to take more than a second? Are you sure you didn't encounter full queues? Maybe instead of using this sleep to distribute the load the ticker should fire 20 times or 100 time per second and have lots of small bursts instead of 1 second bursts? Maybe we should sleep under some load threshold and not under higher ones?

engine/timer.go
```diff
@@ -28,9 +29,15 @@ func Timer(stateI interfaces.IState) {
 			// Leaders running with slightly different minutes in test environments.
 			time.Sleep(time.Duration(s.GetTimeOffset().GetTimeMilli()) * time.Millisecond)
 
-			s.TickerQueue() <- -1 // -1 indicated this is real minute cadence
+			if s.LastSyncTime.Nanoseconds() > tenthPeriod*9/10 && s.Leader {
```
factom-clay: Is measuring the sync time really what we care about? Maybe we should measure the time since this minute started and claim that has to be 90% of a minute instead of this.

engine/loadcreate.go
PaulSnow: I don't know why the sleep limited me to 72 tps.  But removing the sleep allowed me to go to 100 tps.

I am happy with reverting this change.

PR #886, opened 2019-09-30
Fd 1205 thread registry

stackdump:
POC for thread registry

I'm going to try to install it to current codebase - but this is what it looks like sofar


registry/registry.go
```diff
@@ -0,0 +1,93 @@
...
+			Logger: log.ThreadLogger,
+		}
+
+		func() { // add locals to registry
+			StoreManager.Mutex.Lock()
+			defer StoreManager.Mutex.Unlock()
+			StoreManager.Index[gid] = l
+		}()
+
+		<- start // synchronize thread starts
```
factom-clay: Who sends the start signals? Are you faking a wait group? would a  wait group be better? Do you have test code for this ?

factom-clay: I see, you close the start channel and let the nil read start everyone.

log/logger_test.go
```diff
@@ -0,0 +1,10 @@
+package log_test
+
+import (
+	"github.com/FactomProject/factomd/log"
+	"testing"
+)
+
+func TestLogPrintf(t *testing.T) {
+	log.LogPrintf("testing", "unittest %v", "FOO")
```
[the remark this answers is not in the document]
stackdump: I guess I'll assert the file is created

registry/registry.go
stackdump: maybe a wait group would be appropriate
is there a specific scenario you have in mind?

stackdump: I guess I can use one to add a - 'after all threads exit' behavior

stackdump:
rebuilding to use wait groups + will integrate w/ existing codebase

PR #890, opened 2019-10-04
Fd 1207 simplify timer

PaulSnow:
Fix the syncing issue, and make the timer code simple and easy to maintain.


PaulSnow:
Wrong version.

PR #894, opened 2019-10-09
Fd 1205 threadmgr

stackdump:
Thread manager for coordinating the startup of various threads.

https://github.com/FactomProject/factomd/blob/FD-1205_threadmgr/registry/registry_test.go
has examples of how to register Processes & threads

I've testing loading a 'large' db to run a local custom net & intend to continue testing a follower on mainnet

https://github.com/FactomProject/factomd/blob/FD-1205_threadmgr/worker/thread.go#L21-L23
Has some callback hooks that are injected whenever a new thread is spawned

My current thinking is that this mechanism can be extended to inject our global dependencies or to clone copies of data structures that would be local to a single thread.

An added bit of functionality we may not require is the ability to run a separate 'thread lifecycle' in the 'running' context of a parent: called this worker.Thread.Fork() https://github.com/FactomProject/factomd/blob/FD-1205_threadmgr/registry/registry_test.go#L43-L44

^ this would allow for a situation where the init, running, complete, stop wait groups of one 'process' wrap the entire life cycle of another (see example in registry_test.go)


stackdump:
sorry I also merged in the refactor Log + log-all-recoveries branches since those are changes that I would make to be able to inject a thread specific logger

stackdump:
Continuing work as part of FD-1220_hierarchical_obj_names

PR #896, opened 2019-10-14
Fd 1204 make standard metrics

stackdump:
Refactored prometheus usage
added 'label' to all messages

Added Example Guage to Dependant holding - planning to do something similar to all channels/maps (which is the stated goal of this ticket)

Here's a sample of what the metrics endpoint returns
```
queue_size{message="commitchain",name="DependentHolding",package="state",thread="0"} 0
queue_size{message="commitentry",name="DependentHolding",package="state",thread="0"} 244
queue_size{message="revealentry",name="DependentHolding",package="state",thread="0"} 244
```


Emyrk:
I'd just say we should not use a label for ephemeral threads. Only persistent thread, so our label count is bounded.

stackdump:
continuing work as part of https://github.com/FactomProject/factomd/pull/897

@Emyrk Talking w/ Clay about how to label - I think we're going to swap to using names instead of numbers - so we should end up w/ a bounded list.

There is some disparity between running a simulation w/ multiple nodes vs a 'production' configuration where we'd only have one - but the metric system should accommodate either, and I think it will do what you want.

Continuing refinement as part FD-1220_hierarchical_obj_names were we introduce standard names for objects & for threads

Issue #784, opened 2019-07-01
Pending coinbase transactions include non-zero fees 


afenrir:
Pending coinbase transactions retrieved using the `pending-transactions` API method have a non-zero fee:

```
 {
    transactionid: '8153434e365d2d7c80b4906a044c86464d2152a23338bcf8bebd77c17ebd4885',
    status: 'TransactionACK',
    inputs: null,
    outputs: null,
    ecoutputs: null,
    fees: 16500
  }
```

The expected behaviour is for the fee to be 0 or null.

carryforward:
This is odd.  Is this causing a problem.  Would @ThomasMeier like to take a look at this?

carryforward:
I'm wondering if the best course of action is to just not return the coinbase as a pending transaction.  I don't know if the timestamp is set when it is a pending transaction, so a transaction ID would be speculative.

WhoSoup:
> I'm wondering if the best course of action is to just not return the coinbase as a pending transaction.

At the moment, the factom-cli doesn't actually show the coinbase when you type `factom-cli get pendingtransactions` (it only shows transactions that have Inputs), but in all other respects, the coinbase transaction is a transaction and should be shown IMO.

Small PR to just not calculate fees for the coinbase transaction: https://github.com/FactomProject/factomd/pull/898

stackdump:
Confirmed this made it to master branch https://github.com/FactomProject/factomd/blob/178ef7174467dd0b4158091a34a7b61cda88a2e2/state/state.go#L1793

PR #898, opened 2019-10-16
Don't display fees for the pending coinbase transaction

WhoSoup:
closes: #784 

A small fix to just not calculate the fees for the coinbase transaction.

### DIFF
--- a/state/state.go
+++ b/state/state.go
@@ -1720,7 +1720,7 @@ func (s *State) GetPendingTransactions(params interface{}) []interfaces.IPending
 			if pl.DBHeight > currentHeightComplete {
 				cb := pl.State.FactoidState.GetCurrentBlock()
 				ct := cb.GetTransactions()
-				for _, tran := range ct {
+				for i, tran := range ct {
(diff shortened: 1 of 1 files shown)


Emyrk:
Is calculating the fee for any tx incorrect? 
FCT burns into pegnet, for example, would also be incorrect if the `fee` is the fee paid, not the fee expected.

WhoSoup:
> Is calculating the fee for any tx incorrect?
> FCT burns into pegnet, for example, would also be incorrect if the `fee` is the fee paid, not the fee expected.

Yes and no. I think I get what you're saying, that the fee doesn't necessarily represent the amount spent on the transaction. However, in the case of FCT burns the fee is still calculated and paid. Theoretically, when you burn X FCT, you only burn (X-rate) FCT and "spend" the rest as fees. The factom protocol doesn't differentiate between fees and burns. They're essentially the same thing.

If you want to display that information, I would favor keeping "Fees" as is but also adding an "Overpaid" metric

PR #908, opened 2019-10-21
Fd 752

VeenaGondkar:


stackdump:
@VeenaGondkar is this supposed to go into a4 ?

stackdump:
@Emyrk  wants to use named error 

https://golang.org/pkg/errors/#example_New

PR #912, opened 2019-10-29
Fd 1235 prevent election replacing more than half of feds

stackdump:
re-opening PR

TODO: need to add activation height for this change


common/messages/dbstate_test.go
```diff
@@ -350,6 +350,135 @@ func TestSignedDBStateValidate(t *testing.T) {
 	}
 }
 
+// Test that a DBStateMsg.Validate() will return invalid (-1) if the starting federated servers
+// no longer hold a majority in the next block due to replacements
+func TestDBStateValidateReplaceFeds(t *testing.T) {
```
stackdump: This test times out when run locally and on Circle

PR #916, opened 2019-11-08
Fd 1245 rebase

VeenaGondkar:


common/directoryBlock/directoryBlock.go
```diff
@@ -373,6 +373,7 @@ func (b *DirectoryBlock) BuildKeyMerkleRoot() (keyMR interfaces.IHash, err error
 	if err != nil {
 		return nil, err
 	}
+	b.HeaderHash = headerHash
```
stackdump: @VeenaGondkar  seems like the only real code change do we need this?

recovery
```diff
@@ -0,0 +1,2 @@
+2019-11-08 12:13:58.66838948 -0600 CST m=+903.683146929
```
stackdump: it's my fault for leaving this recovery log w/o the '*.txt' extension - maybe I should update the gitignore

PR #921, opened 2019-11-18
Fd 1150 live api test

stackdump:
Merged changes from live API NOTE: 

still expect one more change 
```

[BIF] Sander PostmaToday at 10:08 AM
I do have one addition today (an extra event for when an anchorRecord comes in)
I didn't commit it yes so please let me know whether to use the same git branch or another.
```


stackdump:
Also left a message in FCD discord:

noticed that not all tests from the live api are in '_test' packages
it seems to be out of convience for using some methods in some of the event packages

Unsure if this is really an issue just thought it was notable since this doesn't happen elsewhere in the codebase

stackdump:
Will include changes from this PR https://github.com/FactomProject/factomd/pull/926 for A4

PR #929, opened 2019-11-24
FD-1254 release candidate rolling to master (v6.4.5)

carryforward:
this is the emergency patch that fixes the timing edge case bug for commits.


common/constants/constants.go
```diff
@@ -484,3 +484,4 @@ const (
 //Fast boot save state version (savestate)
 //To be increased whenever the data being saved changes from the last version
 const SaveStateVersion = 13
+const PreBootWindow = 60 // allow an N minute window before boot where messages will be accepted
```
Emyrk: We should revert this to 20

Issue #930, opened 2019-11-25
Infinite loop in Catchup [6.4.5]


WhoSoup:
I'm running 6.4.5 and I noticed that there is an infinite loop in message requests that was causing my testnet follower node to send an excessive amount of data (maxing out my upload basically) and use an excessive amount of cpu. i found the culprit via a cpu trace
the change is this one: <https://github.com/FactomProject/factomd/commit/3f06801a03c1f7935df143ba6fb559a1325a73f7#diff-a6bb2e2c0c0fbd9a9049edb474cb8e01R46>
the state.RequestTimeout doesn't have a default value, so before it would always get set to ~30s, but after that change it's set to 0, causing the second anonymous function to spin out: <https://github.com/FactomProject/factomd/blob/87cc5552c8f0bd5da6dcf584d63714e7a0a5bd72/state/dbStateCatchup.go#L215>, which considers every request to wait for a dblock to be timed out and re-add it as missing, sending out a request to the network.

Suggestion: re-introduce https://github.com/FactomProject/factomd/blob/87cc5552c8f0bd5da6dcf584d63714e7a0a5bd72/state/dbStateCatchup.go#L46-L54 (from 6.4.4) for now

PR #931, opened 2019-11-25
restore old behavior when setting RequestTimeout

stackdump:
https://github.com/FactomProject/factomd/issues/930

@WhoSoup  Identified a bug in this latest RC


factom-clay:
there are type problems with the old code as well. this is fixed in FD-11255 instead.

PR #934, opened 2019-11-26
FD-1243 Release Candidate Confetti (v6.5.0)

carryforward:
this primarily contains grants for the 2019-04 grant round.

This also contains the items from the Rolling release (v6.4.5) which was released as a quick patch to fix a timing issue that was causing problems.


carryforward:
in the merge to develop, this also included the Post-it (v6.4.4) release, that fixed a Missing Message Response bug.

carryforward:
this is the comparison for the difference between this branch and master:
https://github.com/FactomProject/factomd/compare/master...FD-1243_release_candidate_confetti

PR #939, opened 2019-12-10
Fd 1150 live api test

stackdump:
New PR for latest refactor done to live_feed api


database/databaseOverlay/overlay.go
```diff
@@ -124,6 +125,9 @@ type Overlay struct {
 
 	BitcoinAnchorRecordPublicKeys  []interfaces.Verifier
 	EthereumAnchorRecordPublicKeys []interfaces.Verifier
+
+	// We need access to the state to be able emit anchor events
+	parentState events.StateEventServices
```
stackdump: Still looking at all the other changes to understand this - but my first impression is that it's strange to have an events dependency on the DB overlay @Emyrk thoughts?

Emyrk: Yea, the anchors only live in the db. This is a bit strange, and I think to resolve it would require some anchor refactoring.

stackdump: spoke about this - we'll try to do something cleaner during refactor - but will merge this.

Issue #947, opened 2020-01-05
[6.5.0] API Diagnostics index out of range error during elections


WhoSoup:
I'm moving this error from discord/#testnet-chat over to here so it can get some more visibility. It was originally reported by Allen from HashQuark, which is running `v6.5.0-p2p2-beta1-alpine`, though after my initial investigation, I believe the error is likely part of the core code, not the p2p2 beta code or an incompatibility between the two.

The error is an index out of range error, with the full text here: https://gist.github.com/WhoSoup/9722804f43dbef01ff4f0812cf890d5d

The line that fails is \#1588 in the following block: https://github.com/WhoSoup/factomd/blob/387bb5f63a02f1ff82589a17cb8e98c101742378/wsapi/wsapiV2.go#L1581-L1588

It looks like it could be a multithreading problem caused by an "diagnostics" call during an election where the size of available nodes shrinks by one. `state.GetElections()` isn't thread safe and returns only a reference. It is feasible that between lines \#1582 and \#1588, the state of elections changes so that the index is no longer valid.

PaulBernier:
Just want to report that we lost a Leader on mainnet because of that same error. Hope we can raise the prioritization of that bug. Thanks for the initial investigation Who.

stackdump:
Will be released w/ `A4` build

PaulBernier:
Here's actually the full stacktrace: https://pastebin.com/Nh60C8Ak
it seems the node was caught in an infinite loop of such error? Resulting in an out of memory (line 2237) fatal error, causing the actual crash.

stackdump:
That's interesting - wondering if that's behavior from gorilla to try to re-execute

WhoSoup:
> That's interesting - wondering if that's behavior from gorilla to try to re-execute

According to the timestamps, the remote tool calls the diagnostics endpoint roughly every 5 seconds, seems reasonable. Someone with knowledge of the monitoring code should confirm this.

The odd thing is that the error persisted for at least 38 minutes (until crash). This could be an indicator of a problem inside of the elections code and the api error is only a symptom, not the cause.

PaulBernier:
@WhoSoup @stackdump I see a merged PR and a mention of A4 release, can you confirm this is fixed in current Wax release and we can close? Thanks.

WhoSoup:
> I see a merged PR and a mention of A4 release, can you confirm this is fixed in current Wax release and we can close? Thanks.

It is not fixed in Wax.

[...]

PR #948, opened 2020-01-08
Hotfix for diagnostics index out of range error

WhoSoup:
closes: #947 

The elections.Elections struct isn't thread safe, which has caused multiple node crashes for the API "diagnostics" call. It first retrieves the index of the server being elected, then several lines later tries to retrieve the "rounds" count for that server. The "rounds" slice is reset whenever an election is finished and then zero-padded on demand. There are likely two edge cases that can cause a crash:
1. An election finishes right after the index is retrieved, truncating "rounds" to a zero-length slice
2. A new election is started but the "rounds" array hasn't been zero-padded yet

The proper fix would be to create a snapshot of the current "Elections" state via a deep copy to use for the diagnostics call, or provide thread-safe accessor functions for the relevant data, one of which should be done at some point the future.

This fix just makes a copy of the slice and only accesses it if the index is present, using 0 otherwise. For edge case 1., this would result in the loss of the round counter for that call.

### DIFF
--- a/wsapi/wsapiV2.go
+++ b/wsapi/wsapiV2.go
@@ -1585,7 +1585,16 @@ func HandleV2Diagnostics(state interfaces.IState, params interface{}) (interface
 			eInfo.VmIndex = &vm
 			eInfo.FedIndex = &electing
 			eInfo.FedID = e.GetFedID().String()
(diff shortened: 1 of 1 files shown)


carryforward:
ok, this is tracked as FD-1275

PR #950, opened 2020-01-08
Fd 1772 pokemon reporting

factom-clay:


carryforward:
actually ticket FD-1272

PR #951, opened 2020-01-09
Handle API panics more gracefully

WhoSoup:
The golang server is able to handle panics inside of an http request without crashing* but at the moment, panics in are not handled gracefully. It a) prints the stacktrace to stderr, b) sends no output to the client. 

This change adds a panic recovery middleware that addresses both of these.
* If recovering from a panic, write the error to the API log file via `wsLog` like other API errors
* Return an "Internal Error" to the client

\* unless the goroutine causes another goroutine to crash due to multithreading issues or the panic is not recoverable

### DIFF
--- a/wsapi/server.go
+++ b/wsapi/server.go
@@ -113,9 +113,9 @@ func PanicRecovery() Middleware {
 	return func(f http.HandlerFunc) http.HandlerFunc {
 		return func(w http.ResponseWriter, r *http.Request) {
 			defer func() {
-				if r := recover(); r != nil {
+				if rec := recover(); r != nil {
 					trace := debug.Stack()
-					wsLog.Errorf("Recovered from a panic: %v: %s", r, string(trace))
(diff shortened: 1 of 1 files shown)


stackdump (dismissed):
nice choice to make this middleware

wsapi/server.go
```diff
@@ -108,10 +109,26 @@ func IDInjector(server *Server) Middleware {
 	}
 }
 
+func PanicRecovery() Middleware {
+	return func(f http.HandlerFunc) http.HandlerFunc {
+		return func(w http.ResponseWriter, r *http.Request) {
+			defer func() {
+				if r := recover(); r != nil {
```
PaulBernier, line 116: nit but you have 2 variables with the same name that mask one another: `r *http.Request` and `r := recover()`, can you make explicit different variable names?

PaulBernier (dismissed):

PaulBernier (approved):

PR #956, opened 2020-01-17
Fd 1225 sync pubthreaded calls

stackdump:
Added methods for interacting with threaded publisher 
to avoid data race in testing

Did not update the Interfaces - unsure if we want to use this style of interaction during the 'standard' usage of the pubsub package.

### DIFF
--- a/longTest/LeaderModule_test.go
+++ b/longTest/LeaderModule_test.go
@@ -4,12 +4,14 @@ import (
 	"bytes"
 	"github.com/FactomProject/factom"
 	"github.com/FactomProject/factomd/fnode"
+	"github.com/FactomProject/factomd/state"
 	. "github.com/FactomProject/factomd/testHelper"
 	"testing"
 )
 
 //  Currently just tests FCT/EC Commit/Reveal messages to make sure leader is working
 func TestLeaderModule(t *testing.T) {
+	state.EnableLeaderThread = true // enable threaded leader behavior for this test
 
 	encode := func(s string) []byte {
 		b := bytes.Buffer{}
--- a/pubsub/pBase.go
+++ b/pubsub/pBase.go
@@ -63,11 +63,9 @@ func (p *PubBase) Subscribe(subscriber IPubSubscriber) bool {
 }
 
 func (p *PubBase) Write(o interface{}) {
-	p.RLock()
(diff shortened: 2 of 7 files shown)


longTest/LeaderModule_test.go
```diff
@@ -4,12 +4,14 @@ import (
...
 	"github.com/FactomProject/factom"
 	"github.com/FactomProject/factomd/fnode"
+	"github.com/FactomProject/factomd/state"
 	. "github.com/FactomProject/factomd/testHelper"
 	"testing"
 )
 
 //  Currently just tests FCT/EC Commit/Reveal messages to make sure leader is working
 func TestLeaderModule(t *testing.T) {
+	state.EnableLeaderThread = true // enable threaded leader behavior for this test
```
stackdump: unrelated to this pr - but made this change to allow toggle of old/new leader thread behavior during the refactor phase of wax

pubsub/pBase.go
```diff
@@ -63,11 +63,9 @@ func (p *PubBase) Subscribe(subscriber IPubSubscriber) bool {
 }
 
 func (p *PubBase) Write(o interface{}) {
-	p.RLock()
```
stackdump: maybe we still need these?

pubsub/pThreaded.go
```diff
@@ -32,17 +41,16 @@ func (p *PubThreaded) Close() error {
 	return nil
 }
 
+func (p *PubThreaded) WriteWithCallback(o interface{}, callback ...func()) {
+	p.inputs <- o
+}
```
stackdump: cruft will remove

pubsub/pThreaded.go
```diff
@@ -11,9 +11,18 @@ type PubThreaded struct {
...
+	COUNT action = iota + 1
+	SUBSCRIBE
+	UNSUBSCRIBE
+)
+
 type subAction struct {
 	Subscriber IPubSubscriber
-	Subscribe  bool
+	Action     action
+	Reply      chan interface{}
```
stackdump: New optional field to allow caller to wait for response

stackdump: @Emyrk ^

PR #957, opened 2020-01-21
Fd 1175 change to golang 1.13

sanderPostma:
The reason why it started failing with go 1.13 must be a change in behavior in flag.Parse() function which no longer tolerates unspecified command-line parameters, ExitOnError vs ContinueOnError. (Otherwise it would have failed before.)
But this event listener simulator test helper is no longer used after a test refactor last October, so I could just delete the whole thing.


sanderPostma:
The PR also contains some updated files that were generated by protobuf on 1.13. It seems to generate additional some functions.

PR #962, opened 2020-02-04
Reversing API status code

WhoSoup:
This is a compatibility issue that came to light in testing 6.5.1. PR #897 changed the API to return status code 200 for json-rpc errors, but the factom.js client relies on status codes to determine whether or not a request is an error. Functionality is being changed back to preserve backward compatibility.

Reversing this means the FAT/factom client is no longer fully compatible due to https://github.com/AdamSLevy/jsonrpc2/issues/3. However, this is a fairly benign issue that results in FAT/factom to return a json-parsing error rather than an appropriate error.

### DIFF
--- a/wsapi/debugapi.go
+++ b/wsapi/debugapi.go
@@ -30,7 +30,7 @@ func HandleDebug(writer http.ResponseWriter, request *http.Request) {
 	state, err := GetState(request)
 	if err != nil {
 		wsDebugLog.Errorf("failed to extract port from request: %s", err)
-		writer.WriteHeader(http.StatusOK)
+		writer.WriteHeader(http.StatusBadRequest)
 		return
 	}
 
--- a/wsapi/wsapi.go
+++ b/wsapi/wsapi.go
@@ -133,7 +133,7 @@ func returnV1Msg(writer http.ResponseWriter, msg string, success bool) {
 }
 
 func handleV1Error(writer http.ResponseWriter, err *primitives.JSONError) {
-	writer.WriteHeader(http.StatusOK)
+	writer.WriteHeader(http.StatusBadRequest)
 	return
 }
 
--- a/wsapi/wsapiV1.go
+++ b/wsapi/wsapiV1.go
@@ -64,7 +64,7 @@ func checkHttpPasswordOkV1(writer http.ResponseWriter, request *http.Request) bo
 		}
 	} else {
 		wsLog.Errorf("failed to get state from request: %s", err)
-		writer.WriteHeader(http.StatusOK)
+		writer.WriteHeader(http.StatusBadRequest)
 		return false
 	}
 	return true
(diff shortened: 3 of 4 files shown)


PaulBernier:
Just for a bit more context: factom.js is breaking _indirectly_ because the http client it uses, [axios](https://github.com/axios/axios), handle non 2xx response as rejected Promises, which goes to a different path (`catch` instead of `then`), which is a reasonable behavior I think, so any http client having similar handling strategy could also break. The potential for unanticipated breaking goes beyond just factom.js users I think.

carryforward:
ok, this is being tracked with ticket FD-1281

AdamSLevy:
> factom.js is breaking _indirectly_ because the http client it uses, [axios](https://github.com/axios/axios), handle non 2xx response as rejected Promises, which goes to a different path (`catch` instead of `then`), which is a reasonable behavior I think, so any http client having similar handling strategy could also break.

Correct me if I'm wrong but it seems like factom.js is misusing the `catch` path to identify RPC errors. Is axiom also providing the JSON RPC 2.0 parsing? How does the HTTP library matter w.r.t the JSON RPC protocol as used by factom.js?

AdamSLevy:
I would love to see this get merged back, and factom.js be fixed. The FAT/factom package, and fatd that uses it, depends on distinguishing between a network error and a bad lookup. In some cases both errors are treated as fatal, but when it comes to checking to see if an identity is yet populated, the distinction becomes important.

WhoSoup:
> I would love to see this get merged back, and factom.js be fixed.

It seems at this point, we just don't know how many projects rely on factom.js and/or the api status code, so changing it for V2 is unlikely as it would break too many existing projects. This is something now slated for a V3 api, I believe.

PaulBernier:
Axios (which for context is one of the most popular HTTP client with [7M weekly downloads](https://www.npmjs.com/package/axios)) has made the design choice to "route" separately  2xx HTTP response from others. Axios is not aware of RPC, it's just a HTTP client. So as a client I have to use their interface and because factomd returns 4xx response for an error, I have no choice but to handle that in the catch, not up to me to decide. Which is my core point, factomd made a mistake in their RPC2 implementation and clients had no choice but to adapt to this mistake. Of course I totally wish it'd be fixed, but that's just not possible in a backward compatible way. It shows up as a problem in factom.js but could break any other client that worked around factomd current peculiarities...

I have no problem tweaking factom.js to handle both cases and publish a new version, but that would still not make me approve this change implemented this way because of the unknown/unbounded risk of breakage across the entire ecosystem. The correct way to handle this non-backward compatible change is to bump the API version.

AdamSLevy:
Thanks for the clarification. Makes good sense.

Perhaps it would make sense to start an experimental v3 api that is allowed
to make these sorts of changes and is advertised as not stable.

AdamSLevy:
I fixed my jsonrpc2 so that it is not so rigid in enforcing http status codes and always tries to unmarshal the response body. So this won't be an issue for my factom library soon.

PR #965, opened 2020-02-13
Add optional HashType to RawData endpoint

AdamSLevy:
This allows the client to optionally specify the hash type when using the "raw-data" API method. For example, a user typically knows what the hash they are searching for refers to: an entry hash, an EBlock KeyMR, a FCT or EC TX hash. Now they can include that information and reduce needless database look ups. If none is specified, it falls back to searching all of the databases until it finds the hash. This path has also been streamlined.

### DIFF
--- a/wsapi/wsapiV2.go
+++ b/wsapi/wsapiV2.go
@@ -11,7 +11,6 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
-	"reflect"
 	"strings"
 	"time"
 
@@ -713,7 +712,6 @@ func HandleV2RawData(state interfaces.IState, params interface{}) (interface{}, 
 	hashkey := new(HashRequest)
 	err := MapToObject(params, hashkey)
 	if err != nil {
-		panic(reflect.TypeOf(params))
 		return nil, NewInvalidParamsError()
 	}
 
@@ -722,40 +720,55 @@ func HandleV2RawData(state interfaces.IState, params interface{}) (interface{}, 
 		return nil, NewInvalidHashError()
 	}
 
+	dbase := state.GetDB()
 	var block interfaces.BinaryMarshallable
-	var b []byte
-
-	if block, _ = state.FetchECTransactionByHash(h); block != nil {
-		b, _ = block.MarshalBinary()
-	} else if block, _ = state.FetchFactoidTransactionByHash(h); block != nil {
-		b, _ = block.MarshalBinary()
-	} else if block, _ = state.FetchEntryByHash(h); block != nil {
-		b, _ = block.MarshalBinary()
-	}
-
-	if b == nil {
-		dbase := state.GetDB()
-
-		// try to find the block data in db and return the first one found
-		if block, _ = dbase.FetchFBlock(h); block != nil {
-			b, _ = block.MarshalBinary()
-		} else if block, _ = dbase.FetchDBlock(h); block != nil {
-			b, _ = block.MarshalBinary()
-		} else if block, _ = dbase.FetchABlock(h); block != nil {
-			b, _ = block.MarshalBinary()
-		} else if block, _ = dbase.FetchEBlock(h); block != nil {
-			b, _ = block.MarshalBinary()
-		} else if block, _ = dbase.FetchECBlock(h); block != nil {
-			b, _ = block.MarshalBinary()
-		} else if block, _ = dbase.FetchFBlock(h); block != nil {
-			b, _ = block.MarshalBinary()
-		} else if block, _ = dbase.FetchEntry(h); block != nil {
-			b, _ = block.MarshalBinary()
-		} else {
-			return nil, NewObjectNotFoundError()
+	switch hashkey.Type {
+	case HashTypeFCTTx:
+		block, _ = state.FetchFactoidTransactionByHash(h)
+	case HashTypeECTx:
+		block, _ = state.FetchECTransactionByHash(h)
+	case HashTypeEntry:
+		block, _ = state.FetchEntryByHash(h)
+	case HashTypeEBlock:
+		block, _ = dbase.FetchEBlock(h)
+	case HashTypeECBlock:
+		block, _ = dbase.FetchECBlock(h)
+	case HashTypeFBlock:
+		block, _ = dbase.FetchFBlock(h)
+	case HashTypeABlock:
+		block, _ = dbase.FetchABlock(h)
+	case HashTypeDBlock:
+		block, _ = dbase.FetchDBlock(h)
+	default:
+		if block, _ = state.FetchFactoidTransactionByHash(h); block != nil {
+			break
 		}
+		if block, _ = state.FetchECTransactionByHash(h); block != nil {
+			break
+		}
+		if block, _ = state.FetchEntryByHash(h); block != nil {
+			break
+		}
+		if block, _ = dbase.FetchEBlock(h); block != nil {
+			break
+		}
+		if block, _ = dbase.FetchECBlock(h); block != nil {
+			break
+		}
+		if block, _ = dbase.FetchFBlock(h); block != nil {
+			break
+		}
+		if block, _ = dbase.FetchABlock(h); block != nil {
+			break
+		}
+		block, _ = dbase.FetchDBlock(h)
+	}
+	if block == nil {
+		return nil, NewObjectNotFoundError()
 	}
 
+	var b []byte
+	b, _ = block.MarshalBinary()
 	d := new(RawDataResponse)
 	d.Data = hex.EncodeToString(b)
 	return d, nil
--- a/wsapi/hashtype.go
+++ b/wsapi/hashtype.go
@@ -1,0 +1,69 @@
+package wsapi
+
+import (
+	"fmt"
+	"strings"
+)
+
+type HashType int
+
+const (
+	HashTypeUnspecified HashType = iota
+	HashTypeFCTTx
+	HashTypeECTx
+	HashTypeEntry
+	HashTypeEBlock
+	HashTypeECBlock
+	HashTypeFBlock
+	HashTypeABlock
+	HashTypeDBlock
+)
+
+func (ht HashType) String() string {
+	switch ht {
+	case HashTypeFCTTx:
+		return "Factoid Transaction Hash"
+	case HashTypeECTx:
+		return "Entry Credit Transaction Hash"
+	case HashTypeEntry:
+		return "Entry Hash"
+	case HashTypeEBlock:
+		return "Entry Block KeyMR"
+	case HashTypeECBlock:
+		return "Entry Credit Block KeyMR"
+	case HashTypeFBlock:
+		return "Factoid Block KeyMR"
+	case HashTypeABlock:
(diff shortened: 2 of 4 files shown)


AdamSLevy:
I haven't been able to get the tests passing on this but I'm not sure if that's because I don't have the right set up on my system.

wsapi/wsapiV2.go
```diff
@@ -722,40 +720,55 @@ func HandleV2RawData(state interfaces.IState, params interface{}) (interface{},
...
 
+	dbase := state.GetDB()
 	var block interfaces.BinaryMarshallable
-	var b []byte
-
-	if block, _ = state.FetchECTransactionByHash(h); block != nil {
-		b, _ = block.MarshalBinary()
-	} else if block, _ = state.FetchFactoidTransactionByHash(h); block != nil {
-		b, _ = block.MarshalBinary()
-	} else if block, _ = state.FetchEntryByHash(h); block != nil {
```
PaulBernier: you replaced the query `state.FetchEntryByHash(h)` by `dbase.FetchEntry(h)`, do you know the difference? Are you confident that doesn't change the behaviour?

wsapi/wsapiV2.go
```diff
@@ -722,40 +720,55 @@ func HandleV2RawData(state interfaces.IState, params interface{}) (interface{},
...
+	default:
+		if block, _ = state.FetchFactoidTransactionByHash(h); block != nil {
+			break
+		}
+		if block, _ = state.FetchECTransactionByHash(h); block != nil {
+			break
+		}
+		if block, _ = dbase.FetchEntry(h); block != nil {
+			break
+		}
```
PaulBernier: You mentioned that query for Entry is probably the most common so you could optimize the order by putting Entry first. Which I agree. Why did you still put tx queries first?

wsapi/wsapiV2.go
AdamSLevy: I do not know the difference and I am not confident it doesn't change the behavior. I hope someone more knowledgeable can vet that.

AdamSLevy:
> It's a neat addition.
> Can you add 1 or 2 simple test in `wsapiV2_test` that uses your new parameter? There is already a `TestHandleV2GetRaw` there, should be easy enough.

Will do! I'll take care of that tomorrow. I couldn't get the existing tests to pass on my system. I didn't have a database built though...

wsapi/wsapiV2.go
AdamSLevy: When I mentioned that I forgot about tx hashes. I'm not sure which should take precedence: tx hashes or entry hashes. Perhaps entries should come before txs. But I think txs should come before the rest for some gut feeling reason??

wsapi/wsapiV2.go
AdamSLevy: @WhoSoup @carryforward

wsapi/wsapiV2.go
PaulBernier: Ok. I guess it depends on the kind of application, if it's one that work with the monetary side of Factom (like an exchange) or with the data side (pretty much all other uses cases). No big deal but I think I'd still put Entry first.

wsapi/wsapiV2.go
```diff
@@ -722,40 +720,55 @@ func HandleV2RawData(state interfaces.IState, params interface{}) (interface{},
...
+		block, _ = dbase.FetchEBlock(h)
+	case HashTypeECBlock:
+		block, _ = dbase.FetchECBlock(h)
+	case HashTypeFBlock:
+		block, _ = dbase.FetchFBlock(h)
+	case HashTypeABlock:
+		block, _ = dbase.FetchABlock(h)
+	case HashTypeDBlock:
+		block, _ = dbase.FetchDBlock(h)
+	default:
```
WhoSoup: Technically, changing the order in which blocks are searched is a change in functionality though with SHA256 I don't think the likelihood of a conflict existing in the first place is even there. This change should be fine unless I overlooked something obvious.

wsapi/wsapiV2.go
WhoSoup: The amount of people using get-raw in the first place is pretty low, so if they are looking for an optimization they can just specify the hash type.

wsapi/wsapiV2.go
```diff
@@ -722,40 +720,55 @@ func HandleV2RawData(state interfaces.IState, params interface{}) (interface{},
...
+		if block, _ = state.FetchFactoidTransactionByHash(h); block != nil {
+			break
+		}
+		if block, _ = state.FetchECTransactionByHash(h); block != nil {
+			break
+		}
+		if block, _ = dbase.FetchEntry(h); block != nil {
```
WhoSoup: Same as the other comment, this should be the state lookup function.

wsapi/wsapiV2.go
```diff
@@ -713,7 +712,6 @@ func HandleV2RawData(state interfaces.IState, params interface{}) (interface{},
 	hashkey := new(HashRequest)
 	err := MapToObject(params, hashkey)
 	if err != nil {
-		panic(reflect.TypeOf(params))
```
WhoSoup: Technically, this is a change in functionality, though it only affects API requests that send a non-string type for the "Hash" parameter. The change should be fine under the assumption that current clients send valid parameters.

wsapi/wsapiV2.go
AdamSLevy: Thanks for the input. Should both state.FetchEntryByHash and dbase.FetchEntry be called, or just state.FetchEntryByHash?

WhoSoup: Just the state one, it calls dbase.FetchEntry already:
https://github.com/FactomProject/factomd/blob/c0f29f1452a03e36dd1115781ceab99ef2ec3cbf/state/ack.go#L611-L653

wsapi/wsapiV2.go
AdamSLevy: Is panicking really the right response to a bad type? Seems like it should return an invalid params. Is that what panicking does?

If panicking is the appropriate thing, the return statement should be removed as its dead code, which was confusing to me as it was not obvious that the panic wasn't just a remnant of someone's debugging process at one point.

wsapi/wsapiV2.go
AdamSLevy: Does it need both?

wsapi/wsapiV2.go
WhoSoup: No, it's not. It's definitely leftover debug code that ideally shouldn't have been there. It's not a question of right vs wrong, but "changing functionality". Removing bugs can sometimes break programs that depend on those bugs to work.

**I don't think this is one of those cases and the change is fine.**



> Is panicking really the right response to a bad type? Seems like it should return an invalid params. Is that what panicking does?

It's not. panicking will kill the http thread and produce the panic in the console (without impacting the node at all). The client will get no output at all.

I tested it just to make sure and a bad request on the server side looks like this:
```
2020/02/13 08:32:03 http: panic serving x.x.x:57191: map[string]interface {}
goroutine 781658 [running]:
net/http.(*conn).serve.func1(0xc0032d0640)
        /usr/local/go/src/net/http/server.go:1769 +0x139
panic(0xdc2920, 0xccf6e0)
        /usr/local/go/src/runtime/panic.go:522 +0x1b5
github.com/FactomProject/factomd/wsapi.HandleV2RawData(0x1089640, 0xc00016e000, 0xccf6e0, 0xc00b24fce0, 0x0, 0x0, 0x0)
        /go/src/github.com/FactomProject/factomd/wsapi/wsapiV2.go:677 +0x92d
github.com/FactomProject/factomd/wsapi.HandleV2JSONRequest(0x1089640, 0xc00016e000, 0xc002ea49c0, 0x0, 0x0)
        /go/src/github.com/FactomProject/factomd/wsapi/wsapiV2.go:131 +0x676
github.com/FactomProject/factomd/wsapi.HandleV2(0x10644a0, 0xc0005ac1c0, 0xc023c47300)
        /go/src/github.com/FactomProject/factomd/wsapi/wsapiV2.go:65 +0x3dc
github.com/FactomProject/factomd/wsapi.APILogger.func1.1(0x10644a0, 0xc0005ac1c0, 0xc023c47300)
        /go/src/github.com/FactomProject/factomd/wsapi/server.go:96 +0x16d
github.com/FactomProject/factomd/wsapi.IDInjector.func1.1(0x10644a0, 0xc0005ac1c0, 0xc023c47300)
        /go/src/github.com/FactomProject/factomd/wsapi/server.go:106 +0x136
net/http.HandlerFunc.ServeHTTP(0xc00226c3e0, 0x10644a0, 0xc0005ac1c0, 0xc023c47300)
        /usr/local/go/src/net/http/server.go:1995 +0x44
github.com/FactomProject/factomd/vendor/github.com/gorilla/mux.(*Router).ServeHTTP(0xc001c940c0, 0x10644a0, 0xc0005ac1c0, 0xc023c47100)
        /go/src/github.com/FactomProject/factomd/vendor/github.com/gorilla/mux/mux.go:210 +0xe3
net/http.serverHandler.ServeHTTP(0xc003f604e0, 0x10644a0, 0xc0005ac1c0, 0xc023c47100)
        /usr/local/go/src/net/http/server.go:2774 +0xa8
net/http.(*conn).serve(0xc0032d0640, 0x1067760, 0xc002ea4880)
        /usr/local/go/src/net/http/server.go:1878 +0x851
created by net/http.(*Server).Serve
        /usr/local/go/src/net/http/server.go:2884 +0x2f4
```

and on the client side:
```
curl --data-binary "{\"jsonrpc\": \"2.0\", \"id\": 0, \"method\": \"raw-data\", \"params\": {\"hash\": 1}}" -H "content-type:text/plain;" http://x.x.x:8088/v2
curl: (52) Empty reply from server
```

wsapi/wsapiV2.go
WhoSoup: I think so. Writing entries to the database and the processlist are asynchronous, so you might run into race conditions where you have knowledge of the hash but it's not written to the database yet. I'm not 100% certain it's possible, you'd have to do a deep dive of factomd functionality to double check.

AdamSLevy: doesn't hurt to use both I suppose. I guess both should be used under the HashTypeEntry case as well then...

wsapi/wsapiV2.go
AdamSLevy: Thanks for the thorough response.

wsapi/wsapiV2.go
AdamSLevy: but in another comment you said that state.FetchEntryByHash calls dbase.FetchEntry()

WhoSoup: > but in another comment you said that state.FetchEntryByHash calls dbase.FetchEntry()

Yes, it's the third and last place to look for an entry. It tries to get it from the processlist and holding map first.

AdamSLevy:
I just fixed the state vs dbase lookup for Entry Hashes and I added tests where the HashType is specified. 

I can't get the tests to pass probably due to not having the right test database. Maybe someone who has these tests passing can confirm they work. Maybe the CI does that...

PR #966, opened 2020-02-14
Fd 1259 control panel rewrite

lvandenbrink:


lvandenbrink:
The control panel is still work in progress. Currently, not all information is being displayed. To see want still needs to be done, the best thing is to run a factom node and see which parts are missing.
Things that also need to be looked at:
- the control panel is initialized with incorrect leader height. During syncing the displayed progress is invalid.
- the eventUpdates for the index page are being pushed after new every minute. On test-net, with fast increasing minutes, you see the minutes and time not consistently being updated. Probably on main-net the events are pushed slower so it wouldn't be a problem. But it needs to be verified. 
- the process list is being pushed after every change. Only the current process is displayed. Probably the process list should be displayed after every block. Also, the previous process list would be useful. Then the decision need to be made to keep the previous state in the control panel or in the browser.
- data structures for details like summary and process list are published as strings. This was easy as the code was already there. However, it would make more sense to publish a data structure. Probably this would already being pushed for something like the live-feed. The control panel can use a template to serialize the data to a displayable format. Some nice css formatting can be applied to increase the readability. 
- some data is being pushed as an event which can be static information. On the details page the server authorities and identities are being pushed as events. As this doesn't change as frequent this can be loaded on the details page within the template. (similar to NodeName, version and build number for the index page) However, the it can change so the control panel needs to receive an event. As for the My Node information, this is now static but there is a possibility is changes during runtime. 
- the Debug block on the index page should be removed, of course.

PR #967, opened 2020-02-16
Only check for nil-interface when nil is expected

WhoSoup:
In the hunt for pokemon bugs, functions that return an IHash perform the nil interface check. This turns a nil-interface into a nil and also prints a warning if the value is an actual nil. However, there are a couple of functions that explicitly return a "nil" hash right now, which now have a status message printed out.

One of those messages occurs when entries are made to the EC block:
```
MinuteNumber.GetSigHash() returned a nil for IHash. Called from goroutine 2645-/common/entryCreditBlock/minutenumber.go:74
MinuteNumber.GetSigHash() returned a nil for IHash. Called from goroutine 2645-/common/entryCreditBlock/minutenumber.go:74
MinuteNumber.GetSigHash() returned a nil for IHash. Called from goroutine 2645-/common/entryCreditBlock/minutenumber.go:74
...
```

This PR adds a new function `primitives.CheckNilInterface` that allows the return of nil values without a status message and changes the five functions that explicitly return a nil.

(Note: I think I got all the functions, I searched via regex `^\s+defer func\(\) \{ rval.*$[\n\r\s]+return nil`)

### DIFF
--- a/common/entryBlock/entry.go
+++ b/common/entryBlock/entry.go
@@ -143,7 +143,8 @@ func (c *Entry) DatabasePrimaryIndex() (rval interfaces.IHash) {
 
 // DatabaseSecondaryIndex always returns nil (ie, no secondary index)
 func (c *Entry) DatabaseSecondaryIndex() (rval interfaces.IHash) {
-	defer func() { rval = primitives.CheckNilInterface(rval, "Entry.DatabaseSecondaryIndex") }()
+	// reenable if this function is implemented
+	// defer func() { rval = primitives.CheckNil(rval, "Entry.DatabaseSecondaryIndex") }()
 
 	return nil
 }
--- a/common/entryBlock/entry.go
+++ b/common/entryBlock/entry.go
@@ -143,7 +143,7 @@ func (c *Entry) DatabasePrimaryIndex() (rval interfaces.IHash) {
 
 // DatabaseSecondaryIndex always returns nil (ie, no secondary index)
 func (c *Entry) DatabaseSecondaryIndex() (rval interfaces.IHash) {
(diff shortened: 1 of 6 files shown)


common/entryBlock/entry.go
```diff
@@ -143,7 +143,7 @@ func (c *Entry) DatabasePrimaryIndex() (rval interfaces.IHash) {
 
 // DatabaseSecondaryIndex always returns nil (ie, no secondary index)
 func (c *Entry) DatabaseSecondaryIndex() (rval interfaces.IHash) {
-	defer func() { rval = primitives.CheckNil(rval, "Entry.DatabaseSecondaryIndex") }()
+	defer func() { rval = primitives.CheckNilInterface(rval, "Entry.DatabaseSecondaryIndex") }()
 
 	return nil
 }
```
PaulBernier, line 149: I don't understand what this pattern does. What's the purpose of this defer function?

WhoSoup: Nothing in this instance, I just did a find/replace without looking at the placement in-depth.

WhoSoup: Ah, you're right, it seems rather useless.

WhoSoup:
Adjusted the PR to just comment out the nilcheck with a comment to re-enable it if the function is ever implemented. These aren't nil-interface bugs and really shouldn't be reported as such, since they're not implemented.

PaulBernier:
What about merging that in wax branch? Can you rebase to the wax rollup branch?

WhoSoup:
> What about merging that in wax branch? Can you rebase to the wax rollup branch?

Made a new PR since it was far easier than rebasing: #1021

PR #976, opened 2020-03-13
Fd 1283 convert to gomod

stackdump:
In the forked libraries btcutil and btcd there is a bit of complication

I've set the dependencies to commits on specific branches
https://github.com/FactomProject/btcutil/tree/btcec_gomod

https://github.com/FactomProject/btcd/commit/fe42c67cdb3ae36c8bd5dbd3d7ee956c24e62018

https://github.com/FactomProject/web/commit/cff1e06a4e47dac64a68a229ce10e05fd02dbd93

These 3 lilbs could still use more cleanup - some have failing tests.

On a longer term It may be worth thinking about ditching these forks if we can.


stackdump:
I've left the glide files in place for now.
there are several places where they are used in Circle.CI & to load tests using `./test.sh`

PR #978, opened 2020-03-16
Fd 1286 sand rollup

stackdump:
Adds gomod 

TODO: maybe we can incorporate with 'Graph' release 
that will include new networking module.


stackdump:
not going to go this route - instead of this - we are focusing on polishing up wax to be usable/useful

PR #990, opened 2020-04-05
Fixing flaky wsapi unit tests

WhoSoup:
There were two issues at play here:
1. The flaky test was caused by a race condition between the unit test making a http request and the http server starting. This is remedied by the `waitUntilStarted` function, which polls the server every 100ms for the specified duration. The tests now wait for five seconds for the server to start and only then run the tests
2. The wsapi https feature just doesn't work and always returns a 404. This is due to: https://github.com/FactomProject/factomd/blob/ee7ab6b5fb053eefc83ff4f5e223d897f9277faa/wsapi/server.go#L37 This line adds a route for "https://*" but without a handler, shadowing all subsequent routes. The line has been removed again.

### DIFF
--- a/wsapi/wsapi_test.go
+++ b/wsapi/wsapi_test.go
@@ -9,18 +9,36 @@ import (
 	"net/http"
 	"os"
 	"testing"
+	"time"
 
 	"github.com/FactomProject/factomd/common/globals"
 	"github.com/FactomProject/factomd/common/primitives"
-	"github.com/FactomProject/factomd/state"
 	"github.com/FactomProject/factomd/testHelper"
 	. "github.com/FactomProject/factomd/wsapi"
 	"github.com/stretchr/testify/assert"
 )
 
+func waitUntilStarted(t *testing.T, url string, limit time.Duration) {
+	transCfg := &http.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: true}, // ignore expired SSL certificates
+	}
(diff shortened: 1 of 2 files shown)


stackdump:
Not sure we named the next release yet - but this should go into it

wsapi/wsapi_test.go
```diff
@@ -9,18 +9,36 @@ import (
...
 
 	"github.com/FactomProject/factomd/common/globals"
 	"github.com/FactomProject/factomd/common/primitives"
-	"github.com/FactomProject/factomd/state"
 	"github.com/FactomProject/factomd/testHelper"
 	. "github.com/FactomProject/factomd/wsapi"
 	"github.com/stretchr/testify/assert"
 )
 
+func waitUntilStarted(t *testing.T, url string, limit time.Duration) {
```
PaulBernier, line 21: limit -> timeout, slightly better naming

WhoSoup:
> Looks like it'd be worth having in Wax release, what do you think? Can you rebase on wax rollup branch?

Nope, Wax has a different way of starting the test state and has a fix for this problem already.
https://github.com/FactomProject/factomd/blob/f8842079a4c464755cca5f16aaf7bfa428bef9ae/wsapi/wsapi_test.go#L72-L77


This PR was strictly pre-wax.

PaulBernier:
In this case I will just close this PR, I am assuming that Wax will be the next release...

PR #994, opened 2020-04-30
[P2P2] Small updates

WhoSoup:
Both of these surfaced during the testing for my blog:
* Raising the channel capacity allows for higher burst transmissions in the worst case scenario
* V11 broke during high traffic because the code didn't allow for partial reads. switching to io.ReadFull delegates responsibility for partial-reads to golang


PaulBernier:
Closed in favor of https://github.com/FactomProject/factomd/pull/1012 which includes those changes

PR #996, opened 2020-05-06
Make raw data optional in API Requests

WhoSoup:
I was analyzing D/EC blocks for an unrelated issue and noticed that the API will always send both the json data and the raw message, even if you don't need both.  This wastes quite a bit of bandwidth particularly for EC blocks (they're huge due to PegNet). There already is a "get-raw" endpoint for those that only need the raw data but no way to get just the json without the raw.

This changes the following endpoints:
* admin-block
* factoid-block
* entrycredit-block
* dblock-by-height
* ecblock-by-height
* fblock-by-height
* ablock-by-height

("directory-block" itself doesn't even have a raw option, oddly enough)

It adds an optional "noraw" argument to the json call that accepts either "true" or "false". If "noraw" is not present, everything works as before. If "noraw" is true, the API won't return the raw data.

Benchmarked on a LAN copy of factomd, trying to download 3,000 ecblocks  (240650 to 243650):

|  | Time | Size | 
|---|---|---|
|w/ raw|18.6817609s|286.29941082 Mebibyte|
|no raw|13.887978s|160.519980431 Mebibyte|

A ~44% reduction in size and 26% reduction in transfer duration.

### DIFF
--- a/wsapi/wsapiStructs.go
+++ b/wsapi/wsapiStructs.go
@@ -267,6 +267,7 @@ type AddressRequest struct {
 
 type HeightRequest struct {
 	Height int64 `json:"height"`
+	NoRaw  bool  `json:"noraw,omitempty"`
 }
 
 type HeightOrHashRequest struct {
@@ -288,6 +289,7 @@ type HashRequest struct {
 
 type KeyMRRequest struct {
 	KeyMR string `json:"keymr"`
+	NoRaw bool   `json:"noraw,omitempty"`
 }
 
 type KeyRequest struct {
--- a/wsapi/wsapiStructs.go
+++ b/wsapi/wsapiStructs.go
@@ -274,11 +274,6 @@ func (hr *HeightRequest) IncludeRaw() bool {
 	return !hr.NoRaw
 }
 
-type ReplayRequest struct {
-	StartHeight uint32 `json:"startheight"`
-	EndHeight   uint32 `json:"endheight,omitempty"`
-}
-
 type HeightOrHashRequest struct {
 	Height *int64 `json:"height,omitempty"`
 	Hash   string `json:"hash,omitempty"`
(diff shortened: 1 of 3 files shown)


wsapi/wsapiStructs.go
```diff
@@ -273,6 +273,7 @@ type AddressRequest struct {
 
 type HeightRequest struct {
```
PaulSnow, line 274: Consider adding a function:

func (h HeightRequest) IncludeRaw() bool {
    return !h.NoRaw 
}

The problem is that booleans that are the negation of what you need can be very confusing.

wsapi/wsapiStructs.go
```diff
@@ -299,6 +300,7 @@ type HashRequest struct {
 
 type KeyMRRequest struct {
```
PaulSnow, line 301: Same function request

WhoSoup:
Added both suggestions

PaulBernier:
This is looking good, can you rebase the PR against wax rollup @WhoSoup

WhoSoup:
It's rebased. The test that fails is in the p2p package and unrelated to this PR. I'll investigate it in a separate PR.

wsapi/wsapiStructs.go
```diff
@@ -267,6 +267,16 @@ type AddressRequest struct {
...
+}
+
+func (hr *HeightRequest) IncludeRaw() bool {
+	return !hr.NoRaw
+}
+
+type ReplayRequest struct {
+	StartHeight uint32 `json:"startheight"`
+	EndHeight   uint32 `json:"endheight,omitempty"`
 }
```
PaulBernier, line 280: What is this struct doing in this PR? Is that a glitch of the rebase?

WhoSoup: Apparently so. My apologies. It's removed.

PR #999, opened 2020-06-08
[Wax] Fixes and Improvements

WhoSoup:
Using this draft as a collection for a variety of small fixes:

* PubBase wasn't using the RWMutex properly for all accesses of the Subscribers slice
* Pokemon bug in Registry.FindPublisher


modules/pubsub/registry.go
```diff
@@ -52,7 +52,11 @@ func (r *Registry) FindPublisher(path string) IPublisher {
 	r.publock.RLock()
 	defer r.publock.RUnlock()
 
-	return r.Publishers[path]
+	pub, ok := r.Publishers[path]
+	if ok {
+		return pub
+	}
+	return nil
```
PaulBernier, line 59: Why is that the "pokemon bug"? Is it better because if the key is not in the map it returns the "zero value" of the type, and the zero value of an interface is nil anyway, isn't it?

modules/pubsub/pBase.go
```diff
@@ -61,9 +63,11 @@ func (p *PubBase) Subscribe(subscriber IPubSubscriber) bool {
 }
 
 func (p *PubBase) Write(o interface{}) {
+	p.RLock()
 	for i := range p.Subscribers {
 		p.Subscribers[i].write(o)
 	}
+	p.RUnlock()
```
PaulBernier, line 70: I understand you put a _read_ lock because from the Subscribers slice point of view you are only reading, but that means potentially that `Subscribers[i].write()` can be called concurrently for the same Subscriber, and because it's a write operation, could that cause a problem? (i.e is `Subscribers[i].write()` thread-safe?)

WhoSoup: The thread-safety of the subscriber's write function is variable and not the responsibility of the publisher. This gives the flexibility to implement different subscribers and there are examples of both:
* the SubChannel is based on golang channels and channels are already thread-safe, so the subscriber doesn't need its own locks
* the SubConfig is based on a map and has its own lock

modules/pubsub/registry.go
WhoSoup: Hm, that may have been an oversight on my part. I assumed that the map was pointing at structs, in which case this function would have created new structs for unknown keys, but it should be fine for interfaces.

PR #1001, opened 2020-06-10
[Wax] Make worker.Prefix use safe for outside libraries

WhoSoup:
The worker.init() function determines the `Prefix` from the filepath and tries to apply that to all "caller" strings. This makes the worker package unusable outside of the factomd project, since the runtime.Caller file may yield a string that's shorter than `Prefix`. The cost of the additional check should be negligible.

Changes:
* Turn getting the Caller string into an exported function
* Only shorten the string for calls inside factomd
* Add fallback if no caller can be established

### DIFF
--- a/modules/registry/registry.go
+++ b/modules/registry/registry.go
@@ -2,7 +2,6 @@ package registry
 
 import (
 	"fmt"
-	"runtime"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


stackdump:
seems like a good enhancement.

PR #1005, opened 2020-06-24
[Wax] Multithreaded BMV Replay

WhoSoup:
The existing code started two BMV worker threads, each with their own instance of `MsgReplay`. This meant that each worker had their own knowledge of replays, which meant some replays would get through in a very inconsistent manner. This PR modifies the BMV workers to all use a single, shared instance of MsgReplay. In order to accomplish this, MsgReplay had to be made thread-safe. This is done by encapsulating the maps inside a "bucket" struct which provides accessors.


WhoSoup:
Updates:
* Rename Node/msgs to Node/bmv/input
* Rename Node/rest to Node/bmv/output
* Add the state.Blocktime publisher, publishes block start times from MoveStateToHeight()

state/factory.go
```diff
@@ -440,6 +441,9 @@ func (s *State) Initialize(o common.NamedObject, electionFactory interfaces.IEle
 
 	fmt.Printf("The Instance ID of this node is %s\n", s.Salt.String()[:16])
 
+	s.Blocktime = pubsub.PubFactory.Threaded(10).Publish(pubsub.GetPath(s.GetFactomNodeName(), "blocktime"))
```
PaulBernier, line 444: where was this "blocktime" published prior to your change? was it just missing?

modules/bmv/bmv.go
```diff
@@ -48,25 +48,26 @@ type msgPub struct {
...
-	b.rest = pubsub.PubFactory.Threaded(100).Publish(pubsub.GetPath(b.NodeName, "bmv", "rest"), pubsub.PubMultiWrap())
+	b.output = pubsub.PubFactory.Threaded(100).Publish(pubsub.GetPath(b.NodeName, "bmv", "output"), pubsub.PubMultiWrap())
 
-	b.replay = NewMsgReplay(6)
+	b.replay = replay
 	return b
 }
 
 func (b *BasicMessageValidator) Subscribe() {
 	// TODO: Find actual paths
```
PaulBernier, line 69: is this comment still relevant?

modules/bmv/bucket.go
```diff
@@ -0,0 +1,75 @@
...
+	defer b.mtx.Unlock()
+	b.data[key] = time
+}
+
+// Get returns the time attached to the key and true, or zero time and false
+// if the key was not present
+func (b *bucket) Get(key [32]byte) (time.Time, bool) {
+	b.mtx.RLock()
+	defer b.mtx.RUnlock()
+	t, ok := b.data[key]
```
PaulBernier, line 33: don't need "t, ok" as intermediary variables

state/factory.go
PaulBernier: Actually CircleCI is failing with this error:
```
panic: failed to publish: publisher already exists at that path (FNode01/blocktime)
```

PaulBernier: Could it be: https://github.com/WhoSoup/factomd/blob/WhoSoup-bmv_multithreaded/engine/NetworkProcessorNet.go#L75

engine/NetworkProcessorNet.go
```diff
@@ -26,8 +26,8 @@ var _ = fmt.Print
 func NetworkProcessorNet(w *worker.Thread, fnode *fnode.FactomNode) {
 	FromPeerToPeer(w, fnode)
 	stubs(w, fnode)
-	BasicMessageValidation(w, fnode)
-	sort(w, fnode) // TODO: Replace this service entirely
+	BasicMessageValidation(w, fnode, 2) // 2 validators
```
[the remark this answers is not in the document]
WhoSoup: I'm not sure, I went with two because that's what the previous code used. I think it's likely to be premature optimization and just one will do fine. At some point, it will make sense to raise this (perhaps configurable), since this is basically the function that ingests messages from the p2p network and does things like signature checking.

You can ctrl+f for WellFormed() in the code to see that it's not yet entirely clear how heavy the load is.

modules/bmv/bucket.go
WhoSoup: Unfortunately we do, since go's compiler isn't smart enough to figure this one out: https://play.golang.org/p/MC64njwnIT5

state/factory.go
WhoSoup: This seems to be collateral damage from #1037 and explains why matt wasn't able to build this one.

WhoSoup:
Okay i think i managed to place it in the right area this time. it's now part of the "pubregistry" that gets created for every state instance where all the state channels are initialized.

Issue #1006, opened 2020-06-26
Invalid coinbase txs appear in pending-transactions API response with status TransactionACK (pokemon bug)


ilzheev:
We already started investigating it with @WhoSoup 

The case:
1. My Factom Realtime Explorer is polling pending-transactions API with short time interval (seconds).
2. I noticed it catches coinbase invalid factoid txs from time to time (latest affected blocks 250675, 250925, 250950).
3. Invalid coinbase txs appear in pending-transactions API response with status "TransactionACK", but never appeared on the blockchain
4. Invalid coinbase txs appear in ANO payout blocks (% 25 == 0)

Block 250675:
txid 97e18c711c24dd7dc5be9d82722cfd6256575248106bcb4131cb8e11ff0b5534
outputs include payouts to ANOs

Block 250925:
txid 7b028d86dd832acafc041258db1a9a5a7047a82fa828a6670ecded1579e4a850
outputs null

Block 250950:
txid 23a56a843ad0dff7034b8e0c74434798a4a6d32c870ec86b0fa9fb74e2cca255
outputs null

ilzheev:
Hm, another one tx has been caught.
First seen: 27/06 08:49 UTC, equals to block **251126** (next block after ANO payout block).
Outputs null.

<img width="1014" alt="image" src="https://user-images.githubusercontent.com/4902353/85928257-a7fd9e00-b8bc-11ea-8f5c-fd095a5d0616.png">

ilzheev:
I accurately checked timestamps of previous affected txs.

Block 250675:
Timestamp 24/06 05:39 UTC, eq to block **250675**
txid 97e18c711c24dd7dc5be9d82722cfd6256575248106bcb4131cb8e11ff0b5534
**outputs include payouts to ANOs**

~~Block 250925:~~
Timestamp 25/06 23:29 UTC, eq to block **250926**
txid 7b028d86dd832acafc041258db1a9a5a7047a82fa828a6670ecded1579e4a850
**outputs null**

~~Block 250950:~~
Timestamp 26/06 03:39 UTC, eq to block **250951**
txid 23a56a843ad0dff7034b8e0c74434798a4a6d32c870ec86b0fa9fb74e2cca255
**outputs null**

ilzheev:
So, this invalid txs appear on ANO payout block (include ANO payouts in that case) **OR** the block after ANO payout block (and outputs are null).
@WhoSoup @carryforward

ilzheev:
[...]

PR #1009, opened 2020-07-02
[Wax] Stop invalid Coinbase transactions from showing up in pending-transactions

WhoSoup:
This PR is a fix for #1006. This works by adding a flag to FBlocks that track whether or not the coinbase tx has been patched, and in return GetPendingTransactions only returns the coinbase if it has been patched. There are better ways to fix this, such as not creating the coinbase at all until the real timestamp is in, but the code frequently relies on the coinbase transaction existing. I figured this was the least-invasive method that doesn't affect existing behavior.

Changes:
* Add coinbasePatched flag to FBlock and support methods in both the block and the interface
* Add DBHeight to PendingTransaction result that indicates the expected height of the transaction (or 0 for transactions in holding)
* Change the consensus code to use the new FBlock functions instead of blind casts
* Change state.GetPendingTransactions to use string as parameter instead of interface
* Rewrite the GetPendingTransactions to remove completely unnecessary loop over VMs
* Fix console sim input that tried to send just a byte to state.GetPendingTransactions

### DIFF
--- a/state/stateConsensus.go
+++ b/state/stateConsensus.go
@@ -2390,16 +2390,16 @@ func (s *State) ProcessDBSig(dbheight uint32, msg interfaces.IMsg) bool {
 
 			cb := fs.GetCurrentBlock()
 			cbtx := cb.GetTransactions()[0]
-			foo := cbtx.GetTimestamp().GetTimeMilliUInt64()
+			cbTimestamp := cbtx.GetTimestamp().GetTimeMilliUInt64()
 			lts := s.LeaderTimestamp.GetTimeMilliUInt64()
-			s.LogPrintf("dbsig", "ProcessDBSig(): first  cbtx before %d dbsig %d lts %d", foo, dbsMilli, lts)
+			s.LogPrintf("dbsig", "ProcessDBSig(): first  cbtx before %d dbsig %d lts %d", cbTimestamp, dbsMilli, lts)
 
 			s.SetLeaderTimestamp(dbs.Timestamp) // SetLeaderTimestamp also updates the Message Timestamp filter
 
 			uInt64_3 := dbs.GetTimestamp().GetTimeMilliUInt64()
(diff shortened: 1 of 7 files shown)


WhoSoup:
Rebased this to the wax_rollup branch.

state/stateConsensus.go
```diff
@@ -2389,15 +2388,16 @@ func (s *State) ProcessDBSig(dbheight uint32, msg interfaces.IMsg) bool {
 			s.LogPrintf("dbsig", "1st ProcessDBSig(): %10s DBSig dbht %d leaderheight %d VMIndex %d Timestamp %x %d, leadertimestamp = %x %d",
 				s.FactomNodeName, dbs.DBHeight, s.LLeaderHeight, dbs.VMIndex, dbs.GetTimestamp().GetTimeMilli(), dbs.GetTimestamp().GetTimeMilli(), s.LeaderTimestamp.GetTimeMilliUInt64(), s.LeaderTimestamp.GetTimeMilliUInt64())
 
-			cbtx := fs.GetCurrentBlock().(*factoid.FBlock).Transactions[0].(*factoid.Transaction)
-			foo := cbtx.MilliTimestamp
+			cb := fs.GetCurrentBlock()
+			cbtx := cb.GetTransactions()[0]
+			foo := cbtx.GetTimestamp().GetTimeMilliUInt64()
```
PaulBernier, line 2393: is this variable really named foo..... I know it's not you originally but from now on git will attribute it to you, so you may want to rename :P

WhoSoup: > is this variable really named foo..... I know it's not you originally but from now on git will attribute it to you, so you may want to rename :P

Gave them all more appropriate names.

WhoSoup:
> Just to be sure you have tested that change manually by calling the API?

Yeah, @ilzheev and I spent a couple hours working on this problem and fixing it.

PR #1010, opened 2020-07-06
[Wax] Multithreaded ChainHeadFix

WhoSoup:
Motivation: I'm compiling and starting factomd a lot these days and every time I load the MainNet database, there's a ~20 second period where the chain head fix utility runs. I looked at the code and found it to be pretty incomprehensible and also single-threaded. I wanted to see if I could speed up the performance by multithreading the algorithm.

The original algorithm took between 20-21 seconds to run on average. The new code takes 11-12 seconds. The limiting factor now seems to be database i/o.

Description of new algorithm taken from comment:
https://github.com/FactomProject/factomd/blob/bbc5c85fcf05f03f72a4eb76493176c395418ff4/modules/chainheadfix/fix.go#L18-L28

Note: the old algorithm also contained a check for eblock parent block consistency checks. Is this still needed? I was hesitant to outright remove the old code but the functionality isn't enabled in factomd and has never been enabled as far as I can tell.

### DIFF
--- a/modules/chainheadfix/fix.go
+++ b/modules/chainheadfix/fix.go
@@ -38,14 +38,13 @@ type finder struct {
 	headsMtx sync.Mutex
 
 	// multithreaded administrative vars
-	wg     sync.WaitGroup
 	gather chan uint32
 	fix    chan *head
 
 	// counter of how many anomalies were detected
 	errors int64
 
-	// results of the second st ep
+	// results of the second step
 	batchMtx    sync.Mutex
 	batchKeys   []interfaces.IHash
 	batchValues []interfaces.IHash
@@ -82,9 +81,13 @@ func (f *finder) gatherHeads(workers int) error {
 		return nil
 	}
 
-	f.wg.Add(workers)
+	var wg sync.WaitGroup
+	wg.Add(workers)
 	for i := 0; i < workers; i++ {
-		go f.gatherer()
+		go func() {
+			f.gatherer()
+			wg.Done()
+		}()
 	}
 
 	height := head.GetDatabaseHeight()
@@ -93,7 +96,7 @@ func (f *finder) gatherHeads(workers int) error {
 		f.gather <- i
 	}
 	close(f.gather)
-	f.wg.Wait()
+	wg.Wait()
 
 	return nil
 }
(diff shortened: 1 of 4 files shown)


modules/chainheadfix/fix.go
```diff
@@ -0,0 +1,212 @@
...
+
+	// multithreaded administrative vars
+	wg     sync.WaitGroup
+	gather chan uint32
+	fix    chan *head
+
+	// counter of how many anomalies were detected
+	errors int64
+
+	// results of the second st ep
```
PaulBernier, line 48: typo

modules/chainheadfix/fix.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	// empty database
+	if head == nil {
+		return nil
+	}
+
+	f.wg.Add(workers)
+	for i := 0; i < workers; i++ {
+		go f.gatherer()
```
PaulBernier, line 87: it's a bit bewildering that wg.Add and wg.Wait are here, but wg.Done is buried in the gather method. If I want to check the correctness I need to go back and forth between the 2 pieces of code. A more easy approach for future readers would be for me:
```
go func() {
f.gatherer();
f.wg.Done();
}
```
so that all the WG logic is contained in a single place.

PaulBernier: I don't want to argue on any coding style though, we don't have that luxury unfortunately, but I wanted to share my thoughts.

modules/chainheadfix/fix.go
```diff
@@ -0,0 +1,212 @@
...
+				"got":      head.String(),
+				"expected": ch.head.String(),
+				"height":   ch.height,
+			}).Warn("bad chainhead found")
+		}
+	}
+	f.wg.Done()
+}
+
+// FindHeads checkes the databases to find any chains where the chain head
```
PaulBernier, line 199: checkes -> checks

modules/chainheadfix/fix.go
```diff
@@ -0,0 +1,212 @@
...
+
+	log "github.com/sirupsen/logrus"
+
+	"github.com/FactomProject/factomd/common/interfaces"
+	"github.com/FactomProject/factomd/database/databaseOverlay"
+)
+
+// Workers specifies the amounts of worker goroutines used to scan through the database.
+// There is a strong diminishing return for higher values.
+var Workers = 8
```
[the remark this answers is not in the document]
WhoSoup: It's more dependent on the harddrive than the core count, I think. Not sure what would be a good value for this.

modules/chainheadfix/fix.go
WhoSoup: I like that suggestion a lot, now I don't need to have wg be a struct variable.

PR #1012, opened 2020-07-13
[Wax] Updates to P2P2 (Resend filter, minor fixes)

WhoSoup:
Summary from: https://github.com/WhoSoup/factom-p2p/pull/45

> The most significant change would be the introduction of the Resend Filter. This was an idea originally proposed by Clay to be implemented in the old p2p system, though that implementation never worked. The gist is that it takes a certain amount of time for a message to be processed. During that period, nodes will receive the same message a multitude up of times from different nodes. The resend filter will store which nodes sent a specific message and when the time comes to rebroadcast that message, it will only send it to nodes that have not that specific message to us.
> 
> If the processing speed of messages is very fast, this filter has little to no benefit but during times of high traffic/tps, it can cut down on a substantial amount of wasted traffic.
> 
> Other changes:
> 
> * Raising default channel size from 50 to 1000, allowing for higher theoretical burst traffic (discovered during local testing)
> * A race condition for starting the listener
> * Using built-in functions to read data for V11, where before it would fail for fragmented tcp parcels


p2p/configuration.go
```diff
@@ -120,6 +120,15 @@ type Configuration struct {
 	ChannelCapacity uint
 
 	EnablePrometheus bool // Enable prometheus logging. Disable if you run multiple instances
+
+	// PeerResend turns on tracking of application parcels to preventing sending the same
+	// application parcel to peers who already sent it
```
PaulBernier, line 125: preventing -> prevent?
who already sent it -> received it?

p2p/configuration_test.go
```diff
@@ -59,3 +59,19 @@ func TestConfiguration_Check(t *testing.T) {
 		})
 	}
 }
+
+func TestDefaultP2PConfiguration(t *testing.T) {
+	tests := []struct {
+		name  string
+		wantC Configuration
+	}{
+		// TODO: Add test cases.
```
PaulBernier, line 68: hm can you add at least one test case? Otherwise I am not sure it's very useful to add this test at all. But is a test of the default value necessary? What kind of errors would it catch?

p2p/configuration.go
```diff
@@ -163,9 +172,12 @@ func DefaultP2PConfiguration() (c Configuration) {
 	c.ProtocolVersion = 10
 	c.ProtocolVersionMinimum = 9
 
-	c.ChannelCapacity = 50
+	c.ChannelCapacity = 1000
 
 	c.EnablePrometheus = true
+	c.PeerResend = true
```
PaulBernier, line 178: It sounds odd to say PeerResend is true to signify that we do **not** resend messages.
PeerNoResend? PeerMsgDedupe?

p2p/peerResend.go
```diff
@@ -0,0 +1,99 @@
...
+
+type PeerResend struct {
+	mtx      sync.RWMutex
+	buckets  []*PRBucket
+	pc       int
+	interval time.Duration
+	stopper  chan interface{}
+}
+
+type PRBucket struct {
```
PaulBernier, line 17: what does PR stands for? And does this struct needs to be exposed (capitalized)?

p2p/peerResend.go
```diff
@@ -0,0 +1,99 @@
...
+
+func (b *PRBucket) Has(hash [sha1.Size]byte) bool {
+	b.mtx.RLock()
+	defer b.mtx.RUnlock()
+	return b.data[hash]
+}
+
+func (b *PRBucket) Add(hash [sha1.Size]byte) {
+	b.mtx.Lock()
+	defer b.mtx.Unlock()
```
PaulBernier, line 30: you don't need to use defer here, so we can get rid of its overhead, just unlock after the assignment

p2p/peerResend.go
```diff
@@ -0,0 +1,99 @@
...
+	b.data[hash] = true
+}
+
+func NewPeerResend(buckets int, interval time.Duration) *PeerResend {
+	pr := new(PeerResend)
+	pr.interval = interval
```
PaulBernier, line 36: a detail, but it looks like you don't really need to store the value of interval, you could just pass it as an argument to Cleanup and forget about it. Passing it explicitly as an argument of Cleanup has the added advantage of readability as it is clear it's the cleanup interval and not something else

p2p/peerResend.go
```diff
@@ -0,0 +1,99 @@
...
+		if pr.buckets[pr.ptr(i)].Has(hash) {
+			return true
+		}
+	}
+	return false
+}
+
+func (pr *PeerResend) Add(hash [sha1.Size]byte) {
+	pr.mtx.RLock()
+	defer pr.mtx.RUnlock()
```
PaulBernier, line 66: same here, I am not convinced using defer when you don't really need it. My understanding is that it has a minimal overhead, and also slightly decrease readability by having a non sequential order of operations written

p2p/peerResend.go
```diff
@@ -0,0 +1,99 @@
...
+	pr.buckets[pr.pc].Add(hash)
+}
+
+func (pr *PeerResend) Stop() {
+	close(pr.stopper)
```
PaulBernier, line 71: should we protect this with a Do.once? Because calling close on an already closed channel cause a panic I think

p2p/peerResend.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+func (pr *PeerResend) Cleanup() {
```
PaulBernier, line 74: I'd think that method needs to be not exposed

p2p/peerResend.go
```diff
@@ -0,0 +1,99 @@
+package p2p
+
+import (
+	"crypto/sha1"
+	"sync"
+	"time"
+)
+
+type PeerResend struct {
```
PaulBernier, line 9: Not a big fan of that name, it doesn't describe what this class does, it's actually more generic than this name suggest, there is nothing specific about peer or send in this file.
HashTracker? Deduper? HashCache?
I just realized... it is just a kind of LRU cache!

p2p/peerResend.go
WhoSoup: That's true. I'm also unexporting Cleanup as that shouldn't be exported.

p2p/peerResend.go
WhoSoup: Hm that's a tough call. It is idiomatic Golang to have things panic if closed twice, leaving it up to the implementer to make sure they don't close it twice in exchange for lower overhead.

p2p/peerResend.go
WhoSoup: going with "PeerResendFilter" which is a boring name but describes what it does

p2p/configuration_test.go
WhoSoup: This was an auto-generated test block, I removed it.

p2p/configuration.go
WhoSoup: Changed it to "PeerResendFilter" which makes more sense in context

p2p/peerResend.go
WhoSoup: It used to stand for "peer resend" in this case but it should not be exported.

WhoSoup:
Addressed all of the comments except for the Close() one, which I don't think is necessary.

p2p/peerResend.go
PaulBernier: Ok, I see you added a comment on the method, that makes sense

PR #1015, opened 2020-08-06
[Wax] Adding the old ControlPanel

WhoSoup:
This PR drops the old ControlPanel files (with the p2p2 updates) back in. No changes were made to the CP code, only modifying which one is launched and sending the p2p metrics back to the right place.


PaulBernier:
Can you just confirmed that you have built that branch and had verified the basics, like that the control is available, all the different tabs and all the data displayed seems sensible?

WhoSoup:
> Can you just confirmed that you have built that branch and had verified the basics, like that the control is available, all the different tabs and all the data displayed seems sensible?

Yes, I've built it and it more or less works. P2P Connections show up, the tabs work, the processlist shows entries. I have not tested it extensively to see how well the 1/2 pass syncing works, or the internal explorer.

PaulBernier:
Ok I think it's good enough then. I added a [task here](https://github.com/FactomProject/factomd/issues/1017) to ask ANO to test the Control Panel in more details during the testnet release.

PR #1024, opened 2020-08-19
[Wax] Only run circle docker_build for a few branches

WhoSoup:
Right now all PRs fail circle because of docker_build, which is restricted to a few branches. This PR limits the docker_build workflow to only those branches.


PaulSnow:
Thanks WhoSoup!  This has been needed for a while, and had us a bit stumped.

PR #1033, opened 2020-08-22
Some logging adjustments

PaulBernier:
- Restore the flag param `loglvl`
- Adjust and remove some logging


PaulBernier:
Note that this is only some basic minimal improvements, there is a task to rework the whole internal logging: https://github.com/FactomProject/factomd/issues/1036

p2p/controller_connections.go
```diff
@@ -301,7 +301,7 @@ func (c *controller) Dial(ep Endpoint) (*Peer, []Endpoint) {
 	c.logger.Debugf("Dialing to %s", ep)
 	con, err := c.dialer.Dial(ep)
 	if err != nil {
-		c.logger.WithError(err).Infof("Failed to dial to %s", ep)
+		c.logger.WithError(err).Warnf("Failed to dial to %s", ep)
```
WhoSoup, line 304: Not being able to connect to a node is something that is normal, such as when a node is behind NAT/firewall and can't be dialed to. It doesn't necessarily indicate something is going wrong. I was going to say it should be moved to Debug but after writing out this reply, I think a "warning" makes sense in the context. Nothing's wrong but if it happens constantly something might be.

It might get spammy, however.

modules/pubsub/registry.go
```diff
@@ -126,25 +127,25 @@ func globalPublishWith(path string, p IPublisher, wrappers ...IPublisherWrapper)
 	err := globalReg.Register(path, p)
 	if err != nil {
 		tree := globalReg.PrintTree()
-		fmt.Printf("Publish Tree\n%s\n", tree)
+		registryLogger.WithError(err).Errorf("Publish Tree\n%s\n", tree)
```
WhoSoup, line 130: Logrus doesn't need a newline at the end.

modules/pubsub/registry.go
```diff
@@ -126,25 +127,25 @@ func globalPublishWith(path string, p IPublisher, wrappers ...IPublisherWrapper)
...
 		panic(fmt.Sprintf("failed to publish: %s", err.Error()))
 	}
 	return p
 }
 
 func globalPublish(path string, p IPublisher) IPublisher {
-	fmt.Printf("globalPublish: %v\n", path)
+	registryLogger.Debugf("globalPublish: %v\n", path)
```
WhoSoup, line 137: same as above: newline

modules/pubsub/registry.go
```diff
@@ -126,25 +127,25 @@ func globalPublishWith(path string, p IPublisher, wrappers ...IPublisherWrapper)
...
 	err := globalReg.Register(path, p)
 	if err != nil {
 		tree := globalReg.PrintTree()
-		fmt.Printf("Publish Tree\n%s\n", tree)
+		registryLogger.WithError(err).Errorf("Publish Tree\n%s\n", tree)
```
WhoSoup, line 141: same as above: newline

modules/pubsub/registry.go
```diff
@@ -126,25 +127,25 @@ func globalPublishWith(path string, p IPublisher, wrappers ...IPublisherWrapper)
...
 		panic(fmt.Sprintf("failed to publish: %s %s", path, err.Error()))
 	}
 	return p
 }
 
 func globalSubscribe(path string, sub IPubSubscriber) IPubSubscriber {
-	fmt.Printf("globalSubscribe: %v\n", path)
+	registryLogger.Debugf("globalSubscribe: %v\n", path)
```
WhoSoup, line 148: same as above: newline

WhoSoup (changes requested):
Just a minor change

p2p/controller_connections.go
PaulBernier: That was my intention with that Warning, it's not a problem but you should pay attention, maybe blacklist a node that you constantly fail to connect to.

WhoSoup (approved):

PR #1034, opened 2020-08-23
use a Makefile

PaulBernier:
- Replaced build.sh by a Makefile (more conventional, made for builds)
- Added flags trimpath and ldflags -s -w
- Upgraded a bit Docker images
- CLeanup unnecessary files


.gitignore
```diff
@@ -1,5 +1,4 @@
 # Special build files
-engine/overrideversion.go
```
WhoSoup, line 2: Hm, I wasn't actually aware this was in there. I assume that some devs used this to manually set the version instead of using ldflags. Any specific reason for removing it from .gitignore?

WhoSoup (approved):
I'm by no means an expert on build files but I tested that it works and it seems simple enough.

.gitignore
PaulBernier: Just because I removed the ` scripts/prebuild.sh ` that was generating this file, so a priori that file should never show up again. That's seems to me just a hacky way of doing what ldflag can do,

WhoSoup: Ah, some GoLand thing that I'm not using. It's okay with me then but I have no idea if anybody depends on it for whatever reason.

PaulBernier: I have restored this prebuild script and re added to the git ignore

PR #1035, opened 2020-08-23
Cleanup root of the repo

PaulBernier:


WhoSoup (approved):
Not much to say. These files definitely don't belong in the base dir.

PR #1044, opened 2020-08-27
[Wax] EntrySync rework

WhoSoup:
There was a bug in the entry sync process that stuffed the network queue with tens of thousands of messages a second. Personally, I found the entry sync code very hard to understand and it relied on far too many channels and interwoven goroutines.

The rewrite currently is the same algorithm as before but rewritten to be easier to understand and follow. I'm still trying to figure out if something more efficient can be done but haven't come up with anything clever yet. There are some optimizations left, such as a temp cache for checking existence in the database to cut down on the number of database requests.

The throttling works pretty well (tested empirically) and I played around with various numbers. I found that for higher numbers (1000 requests at a time), the average messages per second that we received form the network was between 750 and 800. This number will vary depending on your machine, internet, disk speed, and other factors but I lowered the previous default value of 1000 to 750. This may need to be adjusted in the future or specified by the user.

TODO:
* unit testing
* a full testnet sync


state/entrysync.go
```diff
@@ -0,0 +1,283 @@
...
+		if db == nil { // block not saved yet
+			time.Sleep(time.Second)
+			continue
+		}
+
+		if es.s.DBFinished { // throttle syncing
+			time.Sleep(time.Millisecond * 125)
+		}
+
+		for _, keymr := range db.GetEntryHashes()[3:] { // skip the first 3
```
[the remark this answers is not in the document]
WhoSoup: great point

state/stateConsensus.go
```diff
@@ -1370,42 +1371,21 @@ func (s *State) FollowerExecuteDataResponse(m interfaces.IMsg) {
 
 	switch msg.DataType {
 	case 1: // Data is an entryBlock
-		eblock, ok := msg.DataObject.(interfaces.IEntryBlock)
+		s.Logger.WithFields(logrus.Fields{
```
PaulSnow, line 1374: logrus is a full featured logging package, but comes with very significant performance issues.
http://hackemist.com/logbench/

Clay's logging is very lightweight and is designed to feed into scripts that allow tracking of messages across processes.  Because we end up doing massive amounts of debugging of simulated networks in factomd, this is a huge advantage for development.

For low frequency system logging and exception processing, logrus is fine, but inside processing loops logrus should be avoided.

state/entrysync.go
```diff
@@ -0,0 +1,313 @@
...
+// check routine, for description see EntrySync comment
+func (es *EntrySync) check() {
+	for {
+		select {
+		case <-es.closer:
+			return
+		default:
+		}
+
+		if len(es.eblocks) == 0 { // outside of mutex but should be fine
```
PaulBernier, line 95: Would it be simpler to implement `es.eblocks` as a buffered channel? That way you would not need this polling/wait approach, not need the ebMtx

state/entrysync.go
```diff
@@ -0,0 +1,313 @@
...
+	for _, entryHash := range eblock.GetEntryHashes() {
+		if entryHash.IsMinuteMarker() {
+			continue
+		}
+
+		// see state.UpdateState()
+		update := new(EntryUpdate)
+		update.Hash = entryHash
+		update.Timestamp = ts
+		es.s.UpdateEntryHash <- update
```
PaulBernier, line 268: NOTE: need to check that

state/entrysync.go
```diff
@@ -0,0 +1,313 @@
...
+				es.asks = append(es.asks[:i], es.asks[i+1:]...)
+				i-- // we removed one, index stays the same
+				delete(es.askMap, ask.hash.Fixed())
+				continue
+			}
+
+			// re-ask
+			if time.Since(ask.time) > EntrySyncRetry {
+				request := messages.NewMissingData(primitives.NewTimestampNow(), ask.hash)
+				request.SendOut(es.s, request)
```
PaulBernier, line 159: Just to confirm, this is going to be sent to a single random peer right?

PaulBernier (commented):
You mentioned in your description you'd test a sync on the testnet, did you get a chance to try that out?

common/messages/missingData.go
```diff
@@ -191,8 +191,7 @@ func (m *MissingData) SendResponse(state interfaces.IState) {
 			dataObject = rawObject.(interfaces.IEBEntry)
 			//dataHash = dataObject.(interfaces.IEBEntry).GetHash()
 		case 1: // DataType = eblock
-			dataObject = rawObject.(interfaces.IEntryBlock)
-			//dataHash, _ = dataObject.(interfaces.IEntryBlock).Hash()
+			return // this functionality has been deprecated
```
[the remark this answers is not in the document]
WhoSoup: The entry block doesn't contain entries, only the hashes of the entries, and it's delivered along with the DBState (I think). Requesting missing EBlocks is a functionality that wasn't implemented before my rewrite, I just made it explicit. I'm not sure when it was removed.

state/entrysync.go
```diff
@@ -0,0 +1,313 @@
...
+		time.Sleep(time.Second)
+	}
+}
+
+// returns the maximum possible height we have eblocks for
+func (es *EntrySync) syncMax() uint32 {
+	end := es.s.GetHighestSavedBlk()
+	if es.s.DBStates.ProcessHeight < end {
+		end = es.s.DBStates.ProcessHeight
+	}
```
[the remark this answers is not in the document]
WhoSoup: That's a good question and I'm not sure. I adopted the code from the old entrysync:
https://github.com/FactomProject/factomd/blob/21a1980ab66447018afed2f7d42ac7799239281b/state/entrySyncing.go#L248-L252

ProcessHeight is set when dblocks are processed, but they may not be immediately saved to the database, so it appears there could be a time when a block is considered processed but not saved. I don't know when the reverse would be true, except maybe in edge cases during loading.

state/entrywriter.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"github.com/FactomProject/factomd/database/databaseOverlay"
+)
+
+// WriteEntries consumed the WriteEntry channel and saves entries into the database.
+// Assumes entries are valid.
+// Panics on database errors
+func (s *State) WriteEntries() {
+	for entry := range s.WriteEntry {
```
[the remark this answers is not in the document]
WhoSoup: a graceful shutdown would be closing s.WriteEntry

state/entrysync.go
```diff
@@ -0,0 +1,313 @@
...
+			continue
+		}
+
+		db := es.s.GetDirectoryBlockByHeight(es.position)
+		if db == nil { // block not saved yet
+			time.Sleep(time.Second)
+			continue
+		}
+
+		// legacy logic
```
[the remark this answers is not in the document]
WhoSoup: The intent was that someone who wanted to know why that code was there could look at the old version to see what it did, but you're right that doesn't really help anyone. I included a paraphrase of the original comment.

Sidenote: what do you think about that functionality? It's supposed to interrupt the for{} loop to prevent golang from running it uninterrupted, so that it can focus on loading the database. The original code had a sleep of 1 millisecond, which I'm not convinced does much:
https://github.com/FactomProject/factomd/blob/21a1980ab66447018afed2f7d42ac7799239281b/state/entrySyncing.go#L269-L272

The canonical way to interrupt tight loops would be using `runtime.Gosched()` but it's tricky to benchmark this sort of stuff.

state/entrysync.go
WhoSoup: Using a channel means I'd still have to have an internal loop that sleeps until the eblock is finished, so there's no good way to get rid of polling/wait. It would remove some of the overhead, though. 

I pushed a version that implements this, please let me know what you think.

state/entrysync.go
WhoSoup: It's the following code:
https://github.com/FactomProject/factomd/blob/ba6800ab3ba2296acf82726af2bd64d65c88b7ca/state/state.go#L1375-L1388

state/entrysync.go
WhoSoup: Yup. NewMissingData sets the flag "Peer2Peer" to true:
https://github.com/FactomProject/factomd/blob/ba6800ab3ba2296acf82726af2bd64d65c88b7ca/common/messages/missingData.go#L219

and the P2PProxy takes a "P2P Message with no PeerHash" and sends it to a random peer:
https://github.com/FactomProject/factomd/blob/ba6800ab3ba2296acf82726af2bd64d65c88b7ca/engine/p2pProxy.go#L135-L137

state/entrysync.go
```diff
@@ -0,0 +1,313 @@
...
+func NewEntrySync(s *State) *EntrySync {
+	es := new(EntrySync)
+	es.s = s
+	es.closer = make(chan interface{}, 1)
+	es.askMap = make(map[[32]byte]bool)
+	return es
+}
+
+// Stop initiates shutdown and will stop all associated goroutines
+func (es *EntrySync) Stop() {
```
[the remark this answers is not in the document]
WhoSoup: Added Stop() and closing the entrywriter to the node's shutdown routine:
https://github.com/FactomProject/factomd/blob/18dacb694d07499eb2d40e8195771c41eac10a10/state/state.go#L2407-L2408

WhoSoup:
> You mentioned in your description you'd test a sync on the testnet, did you get a chance to try that out?

It's running, will report back with results.

state/entrysync.go
```diff
@@ -0,0 +1,304 @@
...
+		for {
+			select {
+			case <-es.closer:
+				return
+			default:
+			}
+
+			for i := 0; i < len(eb.missing); i++ {
+				if es.has(eb.missing[i]) {
+					eb.missing[i] = eb.missing[len(eb.missing)-1]
```
PaulBernier, line 104: shouldn't you i-- too? Otherwise `eb.missing[len(eb.missing)-1]` that is now at location `i` won't be evaluated until the next iteration of the outer for loop (which includes a 1s pause)

state/entrysync.go
PaulBernier: Cool, I like the channel use

state/entrysync.go
PaulBernier: ok thanks!

state/entrysync.go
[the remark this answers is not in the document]
PaulBernier: Are you concerned that `runtime.Gosched()` would have more performance impact than the current `time.Sleep(time.Millisecond * 10)`? 
Given the explicit purpose of the sleep here is to give a chance to other goroutines to run, and that's exactly what `runtime.Gosched()` is supposed to do, I'd rather go that way than using sleep as a workaround, no?

state/entrysync.go
WhoSoup: Yup!

state/entrysync.go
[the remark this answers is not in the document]
WhoSoup: I think it's a choice between runtime.Gosched or maybe just delaying execution entirely until the database is finished loading. I'm not sure there's a good reason to have second pass sync running while the node is booting up.

WhoSoup:
I updated the PR with:

Changes:
* Addition of a delayed start, now entrysync waits for the database to finish loading before starting
* Consolidate the two state fields "EntryDBHeightComplete/Processing" and "EntryBlockDBHeightComplete/Processing". They were both doing the exact same thing

Fixes:
* Save the completed height to the database so it can be read the next startup

state/saveAndRestore.go
```diff
@@ -892,45 +874,41 @@ func (ss *SaveState) MarshalBinary() (rval []byte, err error) {
 	if err != nil {
 		return nil, err
 	}
-	err = buf.PushUInt32(ss.EntryBlockDBHeightProcessing)
+
+	// formerly EntryBlockDBHeightProcessing, deprecated
+	err = buf.PushUInt32(0)
```
PaulBernier, line 879: could you leverage save state versioning to just completely get rid of those unused data? it means nodes will need to rebuild their state from 0, but that's not too bad and it's acceptable for Wax which is kind of a "new start" release

PaulBernier: Basically bump SaveStateVersion to 14?

WhoSoup:
I ended up restarting the sync process because I made so many changes (like saving things).

A few hours in:
image

This morning:
image

I also wrote a script that will query the dblock/eblocks via api, then query each entry and verify the hash of the returned data matches the hash in the eblock. I verified the data up to the current sync2 height and everything is there.

WhoSoup:
Seems there are some issues. Checked in on this Monday morning and the first pass was stuck at 117443 (real height: 145k) and the second pass lagged further behind with the message out queue overflowing. Looks like I have some investigating to do.

WhoSoup:
This code has been merged into develop via the wax merge.

PR #1048, opened 2020-09-01
remove isCompilerVersionOK check

PaulBernier:
I upgraded my local setup to go 1.15. As a side effect it looks like it has better static analysis and complained about some string conversion, and for good reason, so I fixed that. Finally, do not disable `vet` for tests.

### DIFF
--- a/engine/factomParams.go
+++ b/engine/factomParams.go
@@ -195,6 +195,9 @@ func isCompilerVersionOK() bool {
 	if strings.Contains(runtime.Version(), "1.14") {
 		goodenough = true
 	}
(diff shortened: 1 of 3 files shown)


WhoSoup:
Thoughts on just removing the whole `isCompilerVersionOK()` functionality? Seems like it's mostly an annoyance for developers and we don't actually guarantee go compatibility down to version 1.7. We'd have to run unit tests for all listed versions for each release and afaik, we don't. ANOs use the docker image, which has a go version specified.

WhoSoup (approved):

PaulBernier:
Agreed, I removed that check

PR #1050, opened 2020-09-01
start addressing vet reported issues

PaulBernier:


common/messages/dbstateMissing_test.go
```diff
@@ -81,35 +81,36 @@ func newDBStateMissing() *DBStateMissing {
...
-					t.Errorf(" Failed with e %d s %d and in %d", e, s, in)
-				}
-
-				if in <= 200 && ne-ns > 200 {
-					t.Errorf(" Failed with e %d s %d and in %d", e, s, in)
-				}
-			}
-		}
-	}
+func TestLimits(t *testing.T) {
```
PaulBernier, line 84: this test used to not run because of the faulty name. After being enabled it failed because of timeouts and/or out of memory. This test doesn't make any sense to test such a trivial function such as NewEnd

simulation/simControl.go
```diff
@@ -1345,7 +1345,6 @@ func StartSimControl(w *worker.Thread, listenTo int, listenStdin bool) {
 				case 'L' == b[0]:
 					if len(b) <= 2 {
 						if cancelheight == -1 {
-							fmt.Errorf("Exp LH.I")
```
WhoSoup, line 1348: Change these to print the error out instead? The rest of the file seems to use either `fmt.Println` or `os.StdErr.WriteString`

util/config_test.go
```diff
@@ -201,7 +202,10 @@ func TestReadConfig(t *testing.T) {
 	if fconfig == nil {
 		t.Error("Empty string ReadConfig() should result in a non-nil config")
 	}
-	fconfig.String()
+
+	if strings.TrimSpace(fconfig.String()) == "" {
+		t.Error("Empty string")
+	}
```
WhoSoup, line 208: Like the one before, the output of this can never be `""`. Unlike the one before, this function can never fail as long as fconfig is not nil, which is checked the line before. 

I think it exists just to get unit test coverage over those lines

WhoSoup (changes requested):
I appreciate the effort to make the orange squiggles go away.

elections/elections.go
```diff
@@ -241,7 +242,7 @@ func (e *Elections) FeedBackStr(v string, fed bool, index int) string {
 }
 
 func (e *Elections) String() string {
-	str := fmt.Sprintf("eee %10s %s  dbht %d\n", e.State.GetFactomNodeName(), e.Name, e.DBHeight)
+	str := fmt.Sprintf("eee %10s %s  dbht %d\n", e.State.GetFactomNodeName(), e.Name.GetName(), e.DBHeight)
```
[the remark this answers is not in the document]
PaulBernier: Changed

common/messages/dbstate_test.go
```diff
@@ -86,7 +87,9 @@ func TestUnmarshalNilDBStateMsg(t *testing.T) {
 
 func TestMarshalUnmarshalDBStateMsg(t *testing.T) {
 	msg := newDBStateMsg()
-	msg.String()
+	if strings.TrimSpace(msg.String()) == "" {
+		t.Error("Empty string result")
+	}
```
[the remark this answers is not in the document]
PaulBernier: Thanks, I like your suggestion

electionsCore/controller/controller_test.go
```diff
@@ -18,29 +18,28 @@ func TestSimpleController(t *testing.T) {
...
-	//ExpMsg(con.RouteLeaderSetLevelMessage(all, 0, all))
-	//fmt.Println(con.GlobalDisplay.String())
-	//
-	//fmt.Println(con.Elections[0].Display.String())
-	//
-	//ExpMsg(con.RouteLeaderLevelMessage(1, 0, all))
-	//ExpMsg(con.RouteLeaderLevelMessage(2, 0, all))
-	//
-	//ExpMsg(con.RouteLeaderLevelMessage(0, 2, all))
+	t.Skip()
```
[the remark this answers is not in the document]
PaulBernier: agree that's not consistent, so will just leave commented

simulation/simControl.go
PaulBernier: Changed for Println

util/config_test.go
PaulBernier: replaced with `_ = ...`, even if it doesn't test anything...

WhoSoup (approved):

PR #1079, opened 2020-10-17
fix display of user provided nodename in CP

PaulBernier:
I noticed my custom nodename was not displayed anymore in the control panel. There are multiple "node name" variable in the code base so that's highly confusing.


WhoSoup (changes requested):
This is kind of weird because the node name (prefix + Node + number) is kind of hard coded and (before this PR) the -nodename parameter does literally nothing (in the old code it was just passed to the control panel like you did). 

It might make more sense to just remove the "nodename" parameter and change "prefix" to "nodename". The control panel includes the prefix already, since it's part of GetFactomNodeName()

PR #1097, opened 2020-10-27
[Nowax] Make RateCalculator thread consistent

WhoSoup:
The Ratecalculator was technically thread safe, as in it didn't crash when accessed by multiple threads, but it wasn't "thread consistent", since it was theoretically possible to update the values while in the middle of loop, making the latter yield inconsistent results.

This change makes the rate calculator use a mutex instead of atomically swapping uints independently so that a single loop execution is always consistent.

### DIFF
--- a/state/rateCalculator.go
+++ b/state/rateCalculator.go
@@ -1,7 +1,7 @@
 package state
 
 import (
-	"sync/atomic"
+	"sync"
 	"time"
 )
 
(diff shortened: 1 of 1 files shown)


PaulBernier (approved):

PaulBernier:
The generally accepted definition of "thread safety" is not that the code doesn't crash in concurrency context but that the correctness of the code is not affected (in particular yields predictable results). So this code was not thread-safe at all... Thanks for fixing that.

PaulBernier (approved):

PR #1099, opened 2020-10-28
remove references to glide (#1046)

PaulBernier:
cherry-pick https://github.com/FactomProject/factomd/commit/b888bfaac9e12099f9e2045c45b27d2222908360

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,2 @@
-# Don't copy the glide vendor directory
 vendor
 .sim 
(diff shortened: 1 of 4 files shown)


WhoSoup:
The test that failed here is from the events module, specifically `eventservices`, which fails because in the event test file, there's an init() function that parses command line flags, which messes with unit tests, which parse their own parameters:

https://github.com/FactomProject/factomd/blob/fd60c1fb7aab41d8c711cf704f1acfbb83a49f60/events/eventservices/eventSim_test.go#L41-L49

At first glance, these tests seem to be intended to run externally with a separate simulated instance of factomd? If so, it might be okay to exclude that directory.

PaulBernier:
Thanks for the initial investigation @WhoSoup . I decided to just rename the method so it doesn't start with Test* and is not considered an automated test anymore.

WhoSoup (approved):

PR #1101, opened 2020-10-28
address some vet issues

PaulBernier:


WhoSoup (approved):
A lot of these look very familiar, my bad on missing that original wax PR.

PR #1109, opened 2020-11-11
add grants 2020-04

PaulBernier:


WhoSoup (approved):
I went through all the application threads and confirmed (via text editor string comparison) that all the addresses match up to the ones listed on the forum. There are a total of 10 successful grants and all 10 are present in the PR.

The activation height (272801) is in ~1952 blocks, which is in ~13 days, 13 hours, 20 minutes, or ([according to my calculations](https://play.golang.org/p/jQC4SNxrrx6)) ~2020-11-24 22:05 UTC

Should give us plenty of time to update.

PR #1111, opened 2020-11-16
Activation Bug Fix + Cleanup

WhoSoup:
Noticed that the /activations code was pretty messy, wanted to clean it up a bit:
* Use sync.Once instead of a bool to calculate network name
* Don't export structs not used or needed outside of package
* Add unit test that double check all the defaults

I discovered a bug where for networks that don't have a custom entry and the default is "disabled", it would always enable them on the first block. Essentially it checked for it and printed out "never activating" but returned true anyway. This fixes that, which has no effect on either MainNet or TestNet. It might have an effect on some unit tests that use a custom network, or any custom network out there.

For that reason, I changed the default to 0 (always active) to keep behavior consistent with how it used to work before.

### DIFF
--- a/activations/activationHeight.go
+++ b/activations/activationHeight.go
@@ -72,12 +72,10 @@ func init() {
 
 // String converts an Activation ID to a name
 func (id ActivationType) String() string {
-
-	n, ok := activationNameMap[id]
-	if !ok {
-		//n = fmt.Sprintf("ActivationId(%v)", id)
+	if n, ok := activationNameMap[id]; ok {
+		return n
 	}
-	return n
+	return fmt.Sprintf("ActivationId(%d)", id)
 }
 
 var netNameOnce sync.Once
@@ -102,7 +100,7 @@ func IsActive(id ActivationType, height int) bool {
 	a, ok := activationMap[id]
 
 	if !ok {
-		fmt.Fprintf(os.Stderr, "Invalid %v (%s)\n", id, id.String())
+		fmt.Fprintf(os.Stderr, "Invalid %d (%s)\n", id, id.String())
 		return false
 	}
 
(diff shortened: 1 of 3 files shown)


state/state.go
```diff
@@ -3076,14 +3076,18 @@ func (s *State) AddToReplayFilter(mask int, hash [32]byte, timestamp interfaces.
...
 	highestCompletedBlk := s.GetHighestCompletedBlk()
 
-	rval := activations.IsActive(id, int(highestCompletedBlk))
+	active := activations.IsActive(id, int(highestCompletedBlk))
 
-	if rval && !s.reportedActivations[id] {
-		s.LogPrintf("executeMsg", "Activating Feature %s at height %v", id.String(), highestCompletedBlk)
+	if !s.reportedActivations[id] {
+		if active {
+			s.LogPrintf("executeMsg", "Activating Feature %s at height %v", id.String(), highestCompletedBlk)
```
PaulBernier, line 3083: isn't highestCompletedBlk  an integer? If so should the string interpolation use %d?

activations/activationHeight.go
```diff
@@ -61,61 +62,59 @@ func init() {
...
 	}
 }
 
 // String converts an Activation ID to a name
 func (id ActivationType) String() string {
 
-	n, ok := ActivationNameMap[id]
+	n, ok := activationNameMap[id]
 	if !ok {
 		//n = fmt.Sprintf("ActivationId(%v)", id)
```
PaulBernier, line 78: would you mind fixing this while you are at it 👀

state/state.go
WhoSoup: %v is the same thing as %d for int, I'll change it for clarity

activations/activationHeight.go
WhoSoup: Done. If you meant something else than what I did, please let me know

PaulBernier (approved):

PR #1113, opened 2020-11-23
Remove useless import hints

WhoSoup:
Not sure why there are so many of these in the codebase but I assume they're leftovers from back when golang tooling wasn't very good? There are a lot of import hints for things like fmt which really shouldn't be necessary.


WhoSoup:
Turns out some of these are load bearing hints

PR #1120, opened 2021-02-17
grants 2021 01

PaulBernier:
- Added grants for 2021 01
- Update version
- The `ValidatorLoop` recovery was swallowing the error making it hard to understand the panic, I added a print statement.


state/validation.go
```diff
@@ -89,7 +89,7 @@ func (s *State) ValidatorLoop() {
 		if r := recover(); r != nil {
 			s.EventService.EmitNodeErrorMessage(eventmessages.NodeMessageCode_GENERAL,
 				"A panic state occurred in ValidatorLoop.", r)
-
+			fmt.Println(r)
```
nklomp, line 92: Any reason to include it in the grant release?

nklomp (commented):
See remark about the println

state/validation.go
PaulBernier: Without it an incorrect grant activation height (which must be divisible by 25 and then added 1) would cause factomd to shut down without telling you the reason why, which was really confusing when I was testing. I included it because it's a minor change directly related to the grant activation height, but that can also go in a separate commit.

state/grants.go
```diff
@@ -1389,6 +1391,69 @@ func GetHardCodedGrants() []HardGrant {
...
+			// WFCT gateway development backpay 2021-01-25 - 2021-02-16
+
+			// De Facto total: 6500 FCT
+			HardGrant{grant2021_01, 6500e8, validateAddress("FA3X3d3ceDYD7v3aBCeEBcvvgiTZNeiFkN8jJ897EbMuBzHt6xhP")},
+			// --------------------------------------------------------
+
+			// --------------------------------------------------------
+			// PaulSnow-Factom Inc-De Facto-Core Committee -- 43000 FCT
+			// Anchor Rework 2021-01-25 - 2021-02-16
+
```
PaulSnow, line 1447: `// Per agreement during question of the grant, the FCT price should be adjusted at the rate`
`// specified by the 144 block average reported by the EC Rate Bot on discord ( #ec-rate-alert )`
`// at a date prior to deploying the grant.  On 2021-2-17 , this rate was $2.57.  This changes the` 
`// tokens issued from 43000 to 23434, or 11712 FCT to Factom Inc. and De Facto`

state/grants.go
```diff
@@ -1389,6 +1391,69 @@ func GetHardCodedGrants() []HardGrant {
...
+			// Factom Inc total: 21000 FCT
+			HardGrant{grant2021_01, 21000e8, validateAddress("FA29PLacx8jVBMZoQrjgES5RBsfooMsWHG5Ep8uUhVmeesGFY4xp")},
```
PaulSnow, line 1449: 21000e8 should be changed to 11712e8

state/grants.go
```diff
@@ -1389,6 +1391,69 @@ func GetHardCodedGrants() []HardGrant {
...
+
+			// De Facto total: 21000 FCT
+			HardGrant{grant2021_01, 21000e8, validateAddress("FA3X3d3ceDYD7v3aBCeEBcvvgiTZNeiFkN8jJ897EbMuBzHt6xhP")},
```
PaulSnow, line 1452: 21000e8 should be changed to 11712e8

state/grants.go
PaulBernier: Thanks @PaulSnow . The math looks slightly off, assuming a new total of 23434, and the core committee part remaining 1000, it gives me: (23434-1000)/2=11217, which is different from your number. It looks like the last 3 digits were suffled, so maybe it's a typo?

PaulSnow (approved):
Thanks for your efforts @PaulBernier

PR #1129, opened 2021-09-30
Fd 1128 code hygiene

Nil425:


PaulSnow (approved):

PaulSnow:
Nil425 shadowed the developer stand up all summer, and took on a small code hygiene task.  
Congratulations on your first commit to an open source project, and looking forward to many more from you!

Emyrk:
Not that it really matters, but you can keep the 1 lines by catching the errors in an accumlator:

```go
	var errs []error
	check := func(_ int, err error) {
		if err != nil {
			errs = append(errs, err)
		}
	}

        check(out.WriteString(fmt.Sprintf("\nFactomd Config")))
        check(out.WriteString(fmt.Sprintf("\n...")))
        check(out.WriteString(fmt.Sprintf("\n...")))

        if len(errs) > 0 {
                return "" // Or return some string that concats all the errors
        }
        return out.String()
```

